PR #13, opened 2016-04-02
Update faq.html

keunes:
- changed name from 'Knowledge Base' to FAQ (as it's named in the app, and what it actually is)
- non-developers won't understand `<foo>`, so replaced that
- added links to store listings in question on ads
- several language corrections

### DIFF
--- a/faq.html
+++ b/faq.html
@@ -87,7 +87,7 @@
               <div class="panel-body">
                 <p>Open the drawer by swiping from the left or tapping the <i class="fa fa-bars"></i> burger icon on the top left corner. Go to <i class="fa fa-plus"></i> <b>Add Podcast</b>.<br />Now, you can either 
                   search the iTunes catalog, browser gpodder.net or add a podcast by URL.</p>
-                  <p>You can also import feeds by opening protocols (such as RSS links) with AntennaPod.</p>
+                  <p>You can also import RSS and Atom feeds as well as URL schemes like <tt>pcast://</tt> and <tt>itpc://</tt> with AntennaPod</p>
                 </div>
               </div>
             </div>
--- a/faq.html
+++ b/faq.html
@@ -87,7 +87,7 @@
               <div class="panel-body">
                 <p>Open the drawer by swiping from the left or tapping the <i class="fa fa-bars"></i> burger icon on the top left corner. Go to <i class="fa fa-plus"></i> <b>Add Podcast</b>.<br />Now, you can either 
                   search the iTunes catalog, browser gpodder.net or add a podcast by URL.</p>
-                  <p>You can also click RSS links (and other protocols) from the browser and select AntennaPod to open them.</p>
+                  <p>You can also import feeds by opening protocols such as RSS with AntennaPod.</p>
                 </div>
               </div>
             </div>
@@ -95,7 +95,7 @@
             <div class="panel panel-default">
               <div class="panel-heading">
                 <h4 class="panel-title">
-                  <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#collapseTwo">The audio is playing way too fast!</a>
+                  <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#collapseTwo">Episodes are playing way too fast!</a>
                 </h4>
               </div>
               <div id="collapseTwo" class="panel-collapse collapse">
@@ -155,7 +155,7 @@
               </div>
               <div id="F-1" class="panel-collapse collapse">
                 <div class="panel-body">
-                  Apparently, you did not install AntennaPod via the Google Play Store. Unfortunately, we can't provide our Flattr API key (and thus Flattr integration) in other builds.
+                  It seems you didn't install Antennapod from the Google Play Store. Your build doesn't contain the necessary credentials for Flattr integration to work.
                 </div>
               </div>
             </div>
@@ -207,7 +207,7 @@
                 </div>
                 <div id="P-3" class="panel-collapse collapse">
                   <div class="panel-body">
-                    Unfortunately, we don't have the time to support other stores besides Google Play. If nothing else, you can get AntennaPod on <a href="https://f-droid.org/repository/browse/?fdfilter=antennapod&fdid=de.danoeh.antennapod">F-Droid</a>.
+                    Unfortunately, we don't have the time to support other stores besides Google Play. If nothing else, you can get AntennaPod on <a href="https://f-droid.org/repository/browse/?fdid=de.danoeh.antennapod">F-Droid</a>.
                   </div>
                 </div>
               </div>
--- a/faq.html
+++ b/faq.html
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta charset="utf-8" />
-  <title>AntennaPod | Knowledge Base</title>
+  <title>AntennaPod | F.A.Q.</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7" crossorigin="anonymous">
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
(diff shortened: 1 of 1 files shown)


faq.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta charset="utf-8" />
-  <title>AntennaPod | Knowledge Base</title>
+  <title>AntennaPod | F.A.Q.</title>
```
mfietz: @TomHennen How common is the usage of this abbreviation in the US? Would, say, grandparents understand?

Also, I'm not sure that FAQ is really better than knowledge base. While most items here are questions, I'm not sure that it will stay this way. This is also meant as an tutorial to get people started.

faq.html
```diff
@@ -70,7 +70,7 @@ <h4 class="panel-title">
             </div>
             <div id="G-4" class="panel-collapse collapse">
               <div class="panel-body">
-                Our builds are completely ad-free and open source! You can even go over to <a href="https://github.com/AntennaPod/AntennaPod">Github</a>
+                Our builds are completely ad-free and open source! Get them at <a href="https://play.google.com/store/apps/details?id=de.danoeh.antennapod">Google Play</a> or <a href="https://f-droid.org/repository/browse/?fdid=de.danoeh.antennapod">F-Droid</a>. You can even go over to <a href="https://github.com/AntennaPod/AntennaPod">Github</a>
```
mfietz: Currently, this is only meant to be used in the app itself, i.e. the user already got the app...

faq.html
```diff
@@ -87,7 +87,7 @@ <h4 class="panel-title">
               <div class="panel-body">
                 <p>Open the drawer by swiping from the left or tapping the <i class="fa fa-bars"></i> burger icon on the top left corner. Go to <i class="fa fa-plus"></i> <b>Add Podcast</b>.<br />Now, you can either 
                   search the iTunes catalog, browser gpodder.net or add a podcast by URL.</p>
-                  <p>AntennaPod can also open various protocols from the browser.</p>
+                  <p>You can also click RSS links (and other protocols) from the browser and select AntennaPod to open them.</p>
```
mfietz: "click... and open" seems a bit verbose.
Maybe "You can also open various file formats like RSS or audio with AntennaPod" (actually, this is not restricted to the browser)?

faq.html
```diff
@@ -95,7 +95,7 @@ <h4 class="panel-title">
             <div class="panel panel-default">
               <div class="panel-heading">
                 <h4 class="panel-title">
-                  <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#collapseTwo">Audio is playing way too fast!</a>
+                  <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#collapseTwo">The audio is playing way too fast!</a>
```
mfietz: Can audio even have an article?
Don't think this is an improvement.

faq.html
```diff
@@ -155,7 +155,7 @@ <h4 class="panel-title">
               </div>
               <div id="F-1" class="panel-collapse collapse">
                 <div class="panel-body">
-                  Apparently, you are using a build not coming from the Google Play Store. These builds don't contain our Flattr API key.
+                  Apparently, you did not install AntennaPod via the Google Play Store. Unfortunately, we can't provide our Flattr API key (and thus Flattr integration) in other builds.
```
mfietz: "You didn't install AntennaPod from the Google Play Store. Your build doesn't contain the necessary credentials for Flattr integration to work."?

faq.html
```diff
@@ -203,12 +202,12 @@ <h4 class="panel-title">
               <div class="panel panel-default">
                 <div class="panel-heading">
                   <h4 class="panel-title">
-                    <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#P-3">Why is AntennaPod not on the &lt;Foo&gt; Store?</a>
+                    <a class="accordion-toggle collapsed" data-toggle="collapse" data-parent="#accordion" href="#P-3">Why is AntennaPod not on Store X or Y?</a>
                   </h4>
                 </div>
                 <div id="P-3" class="panel-collapse collapse">
                   <div class="panel-body">
-                    Frankly, we don't have the time to support other stores besides Google Play. If nothing else, you can get AntennaPod on <a href="https://f-droid.org/repository/browse/?fdfilter=antennapod&fdid=de.danoeh.antennapod">F-Droid</a>.
```
mfietz: Frankly, this really looks like an unnecessary nitpick

faq.html
```diff
@@ -203,12 +202,12 @@ <h4 class="panel-title">
...
+                    Unfortunately, we don't have the time to support other stores besides Google Play. If nothing else, you can get AntennaPod on <a href="https://f-droid.org/repository/browse/?fdfilter=antennapod&fdid=de.danoeh.antennapod">F-Droid</a>.
```
mfietz: Frankly, this really looks like an unnecessary nitpick.
And while you are at it, you could encode the ``&`` with ``&amp;``

faq.html
keunes: Yeah, you mentioned that before. Think that's a great idea, but I guess that should be done on a separate page (otherwise this would be too cluttered/long). Then together they then make up the knowledge base; dunno if you had a vision yet on how the different parts would link together?

faq.html
keunes: I know, but if you're checking this question, I guess a) you probably didn't know about the official channels and b) you probably want to get a different install to get rid of the advertisements.

faq.html
keunes: I'm surprised, but it seems you're right. 'Audio' [apparently](https://dictionary.cambridge.org/dictionary/english/audio) is an adjective (so can't have an article), but it also means that the way you used it isn't correct either (you used it as a noun).

faq.html
keunes: Sorry, I'm a perfectionist, so I often nitpick indeed (while equally often I don't see it as such).
Didn't know `&` should be replaced in a url, but now I'm at it, I will remove the `fdfilter=antennapod` which also makes the `&` unnecessary ;)

faq.html
keunes: Yeah, but this question is about adding feeds, right? (So I wouldn't mention audio)
I guess what we want to say: a) RSS & other protocols are supported; b) feeds can be imported from anywhere on the phone if you select 'AntennaPod' when using these protocols.

Thinking of a next suggestion now.

[I know strictly speaking RSS as such is not a 'protocol', but RSS links are (kind of)]

keunes:
Waiting for @TomHennen's responses btw before doing another commit incorporating @mfietz' suggestions.

faq.html
TomHennen: I like 'unfortunately' better than 'frankly'

TomHennen:
Looks fine to me (after you fix what @mfietz has identified).

faq.html
keunes: So I'll leave FAQ for now, mainly for consistency with the app. Can change back once it's extended.

faq.html
```diff
@@ -87,15 +87,15 @@ <h4 class="panel-title">
               <div class="panel-body">
                 <p>Open the drawer by swiping from the left or tapping the <i class="fa fa-bars"></i> burger icon on the top left corner. Go to <i class="fa fa-plus"></i> <b>Add Podcast</b>.<br />Now, you can either 
                   search the iTunes catalog, browser gpodder.net or add a podcast by URL.</p>
-                  <p>You can also click RSS links (and other protocols) from the browser and select AntennaPod to open them.</p>
+                  <p>You can also import feeds by opening protocols such as RSS with AntennaPod.</p>
```
mfietz: Well, if we are nitpicky: RSS is not a protocol (usually means network protocol, anyway), but a file format.

keunes: That being nitpicky is in my character doesn't mean that you need to copy this behaviour ;)

Concerning 'the right terms': on [this lovely website](http://www.nu.nl/internet/3946198/leukste-podcasts-van-internet-en-zo-abonneer-je.html) I long-press the link to [this RSS feed](http://feeds.feedburner.com/movieinsiders) and chose to 'open with AntennaPod' (or share -> add podcast). From a user perspective I don't download/interact with a file (at least don't experience it as such). Should we still technically correct terms like 'files and intent filters'?

But you're right, on a technical level. Suggestions very much welcome, I've been struggling with this one :)

mfietz: "You can also import RSS and Atom feeds as well as URL schemes like `<tt>pcast://</tt>` and `<tt>itpc://</tt>` with AntennaPod"

keunes:
Updated according to @mfietz' comments. Ready to merge, no?

PR #21, opened 2019-07-12
Using logo without background

ByteHamster:
Shadow of the icon was cut off.

Before/after:


keunes:
The same issue plays up in Google Play. Therefore I'd say it gets fixed together with https://github.com/AntennaPod/AntennaPod/issues/1281

ByteHamster:
I still don't think that we should change the overall appearance of the icon. Updating on Google Play is easy, we can just make the background blue.

keunes:
@ByteHamster What's the hex code of the Google play icon background? I have the impression it's darker than what's currently used on the site as the main colour. I think the two should be aligned (as well as in the app, actually. (and i still think the icon should get a tiny refresh, but that's another discussion so this should be merged)

ByteHamster:
Unfortunately, we are currently quite inconsistent about the color. The app needs some updates in that regard, too (different colors in different pages).

Issue #24, opened 2019-07-21
Update FAQ to represent new Flattr situation


keunes:
Feature removed from code, but still in many installs. Add a question for now for sth along the lines of 'where did Flattr go?', or leave as is and remove alltogether after some time.
https://github.com/AntennaPod/AntennaPod/pull/3266

orionlee:
Home page (https://antennapod.org/) needs to be updated as well.

keunes:
@orionlee Good point. I just wanted to remove, but it's actually listed in the 'main features' section. It would be nice to replace it with something else. Any suggestions?

orionlee:
Open source is a possibility, though it's already mentioned right at the top.

Another is ad free and/or privacy (as little tracking as possible).

keunes:
Yeap, I like it. Now next question: which icon? Here some options. I would suggest going for `eye-slash' or 'user-shield'.

https://fontawesome.com/icons/mask?style=solid
https://fontawesome.com/icons/lock?style=solid
https://fontawesome.com/icons/user-shield?style=solid
https://fontawesome.com/icons/user-lock?style=solid
https://fontawesome.com/icons/shield-alt?style=solid
https://fontawesome.com/icons/user-secret?style=solid
https://fontawesome.com/icons/eye-slash?style=solid

keunes:
Ok, I think I'd go for the user with shield:
<img alt="https://fontawesome.com/icons/user-shield?style=solid" src=https://user-images.githubusercontent.com/11229646/82379159-bbe4f300-9a26-11ea-8799-ef3df24b3d93.png width=200px>

And the following:
**Ad-free**
Protect your privacy by keeping out ads and other tracking mechanisms

@orionlee @ByteHamster?

ByteHamster:
> keeping out ads

This sounds like it includes an ad blocker. Maybe something like "not showing ads"?

keunes:
That's a good point. Not sure about 'not showing ads' either. What about the following?

**Ad-free**
Protect your privacy with an ad-free app without tracking mechanisms

OR
Protect your privacy - enjoy the app ad-free and without tracking mechanisms

Protect your privacy with our app free of ads and other tracking mechanisms

Protect you privacy and enjoy the app ad-free and without tracking mechanisms

ByteHamster:
> Ad-free
> Protect your privacy with an ad-free app without tracking mechanisms

Sounds good to me :+1:

[...]

PR #26, opened 2019-10-25
Removed flattr from FAQ

ByteHamster:
We no longer support Flattr


keunes:
It was still in the FAQ for those who are on older editions - can you assess how many people could still be seeing Flatter?
Related: https://github.com/AntennaPod/AntennaPod/pull/3266 and #24

ByteHamster:
I never thought about that. We have 132k active installs. 116k use version 1.7.3b or 1.8.0-RC. Let's wait a bit more before merging this PR.

ByteHamster:
I think enough time has passed to merge this. We even have an "official" statement by Flattr ;)

image

PR #31, opened 2020-03-16
Remove third party deps and host them instead

stevenroose:
Currently the website pulls in a bunch of third party deps making it fragile for unavailability of them and leaking usage data to these third parties.

screenshot-2020-03-16-14-26-57


ByteHamster:
Thanks! Good point. Just a little question: Where did you get the ttf files from? They seem to be different from the ones that can be downloaded on Google Fonts.

Lato is an open font, so they could be renamed to not contain the word Google :)

stevenroose:
I fetched them from where they were actually fetched.

You can `wget` or `curl` the remote source that I removed: `//fonts.googleapis.com/css?family=Lato:300,400,900`.

In that CSS file there are the links to the ttf files that are pulled in. So I used `wget` to get those files in and then renamed them.

Feel free to suggest another naming of the files. Or for me to use other sources for the files.

stevenroose:
If you prefer I can just remove all the google naming.

ByteHamster:
> You can wget or curl the remote source that I removed: //fonts.googleapis.com/css?family=Lato:300,400,900.

Interesting, if I open that with a browser instead of curl, I get only woff files.

> If you prefer I can just remove all the google naming.

That would be great :)

stevenroose:
K this probably cost me more time than I liked. But I noticed the faq.html and privacy.html pages and I made it work for them also. I opted to include the entire font-awesome and bootstrap dist folders (except scss and less files) because otherwise the icons don't work.

stevenroose:
done

stevenroose:
Haha, that's just boring website maintenance. I thank you for such an awesome application! Which reminds me that I've been wanting to put out a feature request for a button to slow down the playback speed so I don't have to cycle all the way through my enabled playback speeds :)

PR #32, opened 2020-05-19
Remove mailto: link

keunes:
In the past we had an issue where users contacted the community list from the app via the email address and consequently did not get an update on any replies on their tickets. This was circumvented by providing a link to the group instead. Same issue does and solution should apply to the website.


tonytamsf:
Nice

PR #39, opened 2020-11-23
[index.html] remove www from github.com address

arthurlutz:


keunes:
Hi @arthurlutz,

We're very close to releasing our new website - actually we will tomorrow (Sunday, CET). So we'll close this issue. But thanks for chipping in to help us! If you want to have a look at our new website, check out #40.

Best,
Keunes

PR #40, opened 2020-11-26
New website

ByteHamster:


_includes/head.html
```diff
@@ -0,0 +1,51 @@
...
+  <meta name="msapplication-TileColor" content="#2a854e">
+  <meta name="theme-color" content="#2a854e">
+  <link rel="shortcut icon" href="{{ site.baseurl }}/favicon.ico" type="image/x-icon" />
+  <link rel="stylesheet" href="{{ site.baseurl_root }}/assets/css/main.css">
+  <link rel="stylesheet" href="{{ site.baseurl_root }}/assets/css/all.css">
+  <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: site.baseurl | prepend: site.url }}">
+  <!-- JavaScript -->
+  <script src="{{ site.baseurl_root }}/assets/js/jquery.min.js"></script>
+  <script src="{{ site.baseurl_root }}/assets/js/bootstrap.bundle.min.js"></script>
+  <script src="{{ site.baseurl_root }}/assets/js/search.min.js"></script>
```
Rotzbua, line 50:
```suggestion
  <script defer src="{{ site.baseurl_root }}/assets/js/search.min.js"></script>
```
improves load speed. jquery and bootstrap could not be optimized because you use inline javascript. maybe move the inline javascript to extra file

Rotzbua (commented):
Add a `.editorconfig` to avoid different code styles. Github and many IDEs uses this file .

keunes:
Thanks for the review @Rotzbua ! I'll update the links where we can use https, hope that @ByteHamster and @AnXh3L0  can have a look at your other comments.

.github/workflows/main.yml
```diff
@@ -0,0 +1,25 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - new-website
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
Rotzbua, line 10:
```suggestion
    runs-on: ubuntu-20.04
```
better pinned, to avoid breaks by external version increase

_i18n/en/about.md
```diff
@@ -0,0 +1,34 @@
...
+
+# We hear you: community & accessibility
+AntennaPod pushes the content in your ears, but we like to hear from you as well! Our community is open to your feedback and your thoughts. As humans we can disagree, but do so respectfully. And we do our best to make the app friendly for everyone. That's why it is possible to hide automatic suggestions and easy to navigate around the app through talkback.
+
+# Est. 2011: the AntennaPod project today
+The [first bit of code](https://github.com/AntennaPod/AntennaPod/commit/c9283f09dced6f156e13675ef4c13ebeb20cb9e5) of AntennaPod was published in 2011 by Daniel Oeh, the original creator of AntennaPod. Since then, the app has improved and our community grown.
+
+Today, AntennaPod brings together a large community of podcast addicts and open source enthusiasts. The app is installed on over 179.000 active devices (a phone or tablet that's been turned on at least once the past 30 days) via Google play. And we suspect thousands of users have installed the app via F-Droid, the privacy-friendly app store.
```
ByteHamster, line 30: I would write "over 180.000". The rounded number looks better (and we are now at 181.000). Should this use a comma instead of a dot (`180,000`)?

_i18n/en/about.md
```diff
@@ -0,0 +1,34 @@
...
+
+The whole project is maintained by a loose collective of volunteers from around the world. There is no official organisation behind it. There are about 3 people who regularly contribute code, with [many more](https://github.com/AntennaPod/AntennaPod/graphs/contributors) chipping in on an occasional basis. Our forum has over 190 members and each month attracts well over 300 users. The project is currently lead by @ByteHamster and @Keunes.
```
ByteHamster, line 32: > each month attracts well over 300 users

To me, this sounds like the number of users increases by 300 each month.

ByteHamster:
Thanks for the review @Rotzbua. I applied your suggestions.

_i18n/en/about.md
keunes: `and each month has well over 300 visitors`?

PR #61, opened 2021-02-24
Add optional dark mode

### LINKED ISSUE #60: Add dark theme
We have a media query for that https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

I can do it.

vitaly-zdanevich:
Closes #60

### DIFF
--- a/_sass/_dark.scss
+++ b/_sass/_dark.scss
@@ -1,0 +1,91 @@
+@media (prefers-color-scheme: dark) {
+    :root {
+        scrollbar-color: #333 #000;
+    }
+
+    body {
+        color: #bbb;
+        background-color: #111;
+    }
+
+    .content h1 {
+        color: #ddd;
+    }
+
+    .nav-item:hover {
+        background: #666;
+    }
+    .nav-item.active {
+        background-color: #444;
+    }
+    .navbar-collapse > .navbar-nav > .nav-item > .nav-link {
+        color: #bbb;
+    }
+    .navbar-collapse > .navbar-nav > .nav-item > .nav-link:hover {
+        color: #fff;
+    }
+
+    .dropdown-toggle:after {
+        filter: invert(0.8);
+    }
+
+    .navbar-toggler {
+        filter: invert(1);
+    }
+
+    img.phone,
+    img.half-phone {
+        border-color: #333;
+    }
+
+    .post-content p {
+        color: #ddd;
+    }
+    .project-card,
+    .highlight-card {
+        color: var(--primary);
+        background-color: #000;
+    }
+    hr {
+        border: none;
+    }
+
+    path {
+        fill: #223;
+    }
+
+    #search-input {
+        color: #fff;
+    }
+
+    .copyright {
+        color: #bbb;
+    }
+    .copyright a:hover {
+        color: #fff;
+    }
+
+    .post-header a,
+    .post-content a {
+        color: #bbf;
+    }
+
+    .overlay {
+        background-color: #333;
+    }
+    .overlay a {
+        color: #bbb;
+    }
+    .overlay a:hover {
+       color: #000 !important;
+    }
+
+    .background-gray-100 {
+        background-color: #223 !important;
+    }
+    .color-gray-800,
+    .color-gray-900 {
+        color: #ddd !important;
+    }
+
+}
--- a/_sass/_dark.scss
+++ b/_sass/_dark.scss
@@ -8,7 +8,8 @@
         background-color: #111;
     }
 
-    .content h1 {
+    .content h1,
+    .content h4 {
         color: #ddd;
     }
 
--- a/_sass/_dark.scss
+++ b/_sass/_dark.scss
@@ -15,12 +15,15 @@
     .nav-item:hover {
         background: #666;
     }
+
     .nav-item.active {
         background-color: #444;
     }
+
     .navbar-collapse > .navbar-nav > .nav-item > .nav-link {
         color: #bbb;
     }
+
     .navbar-collapse > .navbar-nav > .nav-item > .nav-link:hover {
         color: #fff;
     }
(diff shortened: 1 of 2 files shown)


ByteHamster:
Thanks for working on this. I made a few screenshots of pages that still need some improvements. In general, please try to avoid `!important` statements. For color schemes, please have a look at our branding repo: https://github.com/AntennaPod/Branding



- Download button has unreadable font color
- Download button no longer uses the brand color
- Menu items do not use a brand color
- Phone border is too bright



- Copyright unreadable



- Headings unreadable
- Menu item highlight color does not match the color scheme



- Mobile menu not styled
- Mobile menu icon invisible
- Link font does not match the color scheme

vitaly-zdanevich:
Why not `!important`? Against them I can move out **light** code into `@media (prefers-color-scheme: light)`

vitaly-zdanevich:
> Download button no longer uses the brand color

No, its still the same `#177ef0` from `.hero`.

ByteHamster:
> Why not !important?

Usually, values that are specified further down in the css file or are more specific overwrite other values automatically. If the values are not overridden and you need `!important`, chances are high you forgot a use-case that is still not styled properly.

ByteHamster:
> No, its still the same `#177ef0` from .hero.

Woah, interesting. In the light theme, logo color and button color look the same to me (even if you are right - it's different colors). In the dark theme, they look noticeably different.

vitaly-zdanevich:
> > No, its still the same `#177ef0` from .hero.
> 
> Woah, interesting. In the light theme, logo color and button color look the same to me (even if you are right - it's different colors). In the dark theme, they look noticeably different.

There balls are the same color :)
image

vitaly-zdanevich:
> Headings unreadable

What do you mean? Where?

vitaly-zdanevich:
> > Why not !important?
> 
> Usually, values that are specified further down in the css file or are more specific overwrite other values automatically. If the values are not overridden and you need `!important`, chances are high you forgot a use-case that is still not styled properly.

Now `!important` is only for two selectors from Bootstrap.

vitaly-zdanevich:
Fixed your feedback, please check.

I think we need dark screenshots for the main - I can do it with any podcast thumbnails inside? And maybe proportion of my phone is not identical to your..

keunes:
@vitaly-zdanevich Please see here for the screenshots: https://github.com/AntennaPod/Branding/tree/master/Screenshots
As written there, the database available there should be used. Valid point about screen size - last time I offered to make screenshots it was decided that it's better to keep existing screen size, so ByteHamster did it then. I'll let him confirm this case.

ByteHamster:
To be honest, I don't think we should create dedicated screenshots for the dark theme. That doubles the amount of work to keep them up-to-date and translated into all languages.

vitaly-zdanevich:
> Please see here for the screenshots: https://github.com/AntennaPod/Branding/tree/master/Screenshots

I found dark screenshot from here  

https://github.com/AntennaPod/Branding/blob/master/Screenshots/raw/en-US/03b.png

vitaly-zdanevich:
So what about CSS? Its done?

vitaly-zdanevich:
And this is strange that now on main we have two identical screenshots:

image

keunes:
> And this is strange that now on main we have two identical screenshots

I think it was added as a placeholder. The new site went live late last year, but many of the [ideas](https://forum.antennapod.org/t/sitemap-for-the-new-website/240?u=keunes) didn't get to be implemented (yet) – I think this is one of the points that was left also.

ByteHamster:
> I found dark screenshot from here

Please just keep the light one. When we later add more screenshots, this will be a huge amount of work to keep both themes updated.

> I think it was added as a placeholder.

Yeah, I just used the same image again because I wanted to get the new website design with the new logo out as quickly as possible. No one has complained yet, so I guess it's okay to keep using the same image until I have some more time :smile: 

> Headings unreadable
>> What do you mean? Where?

The contrast of the headings is really low. On my device, they are basically unreadable when there is light in the room. Please ensure WCAG AAA contrast for all UI elements (see https://webaim.org/resources/contrastchecker/).

image

> So what about CSS? Its done?

- The dropdown button of the language selector should be inverted, too
- The link color of blog post authors is unreadable
- Links and icons on the documentation pages are really hard to read
- Maybe it's just my monitor but I can barely see the difference between the header background color and the page background color (was light blue before and is now full black)
- I think the contrast of the main fonts is a bit low. I would change the main font color to at least `#ddd`.

vitaly-zdanevich:
About `About` in the header: in Firefox devtools I see it **AAA** by WCAG.

image

vitaly-zdanevich:
Done all your feedback, please check.

ByteHamster:
I like the new header background color :+1:

> About About in the header: in Firefox devtools I see it AAA by WCAG.

I meant the title of the page that is displayed when clicking the menu item.

The normal text of the about page should use `#ddd` instead of `#bbb` like on the main page.

vitaly-zdanevich:
Done:

image

ByteHamster:
> The normal text of the about page should use `#ddd` instead of `#bbb` like on the main page.

This is still missing. Also, the contrast of the headings on the "contribute" page are pretty low.

I think it would be great if you could move the dark theme code to its own css file (and use `@include` in the existing file). Helps to make the code a bit more modular.

vitaly-zdanevich:
> the contrast of the headings on the "contribute" page are pretty low

Where?

image

ByteHamster:
> Where?

The headings of the cards. I can barely read them (or is this just my monitor?). See image below.

image

Thanks for moving to a new file. For a consistent code style, could you please ensure that there always is an empty line between two css blocks in your new code?

```
x {
    ...
}
HERE
y {
    ...
}
```

vitaly-zdanevich:
Done, please check.

ByteHamster:
Sorry for the delay. I took a few days off AntennaPod development. Thanks for formatting the code and moving it. There still are headings with low contrast left (on the documentation page). See the screenshot below.

image

vitaly-zdanevich:
Done:

image

ByteHamster:
Thanks! Looks good to me now.

PR #66, opened 2021-04-16
Update Beta documentation

keunes:
Add explanation to why there is no beta on F-Droid; add max-width for Google beta tester image.

### DIFF
--- a/_documentation/general/beta.md
+++ b/_documentation/general/beta.md
@@ -17,7 +17,7 @@ Every new version of AntennaPod is sent to our Beta Testers via the Google Play 
 
 So you want to have the latest and greatest? That's certainly possible! Simply visit the AntennaPod listing in the Play store and go to "Become a beta tester" at the bottom of the page. Or, if you're at your computer, you can go to [this url](https://play.google.com/apps/testing/de.danoeh.antennapod).
 
-<img src="/assets/images/documentation/project-beta.png" alt="Become a beta tester!" style="max-width:400px;">
+![Become a beta tester!](/assets/images/documentation/project-beta.png){: width="400" }
 
 ## F-Droid
 We do not currently offer beta versions on the F-Droid store.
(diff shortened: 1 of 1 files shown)


_documentation/general/beta.md
```diff
@@ -17,7 +17,9 @@ Every new version of AntennaPod is sent to our Beta Testers via the Google Play
 
 So you want to have the latest and greatest? That's certainly possible! Simply visit the AntennaPod listing in the Play store and go to "Become a beta tester" at the bottom of the page. Or, if you're at your computer, you can go to [this url](https://play.google.com/apps/testing/de.danoeh.antennapod).
 
-![Become a beta tester!](/assets/images/documentation/project-beta.png)
+<img src="/assets/images/documentation/project-beta.png" alt="Become a beta tester!" style="max-width:400px;">
```
ByteHamster, line 20: The website theme already sets the `max-width` to 100% of the page width. If you now set it to 400px, that is overwritten and can lead to an overflow on phones (see screenshot below when I just changed the CSS on the "Inspect element" page of the browser). Assigning the `width` CSS property directly would be more reliable. Then it is 400px by default but might also be smaller when the page is smaller.

image

keunes: Whoops. Yes, thank you. Will update it.

keunes:
Ok to merge?

ByteHamster:
Looks like the new syntax you use doesn't work. It sets the `width` attribute of the image tag instead of using CSS. Apparently, the `width` attribute is "stronger" than CSS, so `max-width` is not applied.

keunes:
Ok, I looked for Jekyll kramdown and this came up ¯\_(ツ)_/¯

Next attempt. And I should really learn git, so I can check this stuff locally.

ByteHamster (approved):

ByteHamster:
> And I should really learn git, so I can check this stuff locally.

Do you know the GitHub desktop app? It might be easier than learning the git command line. If you want to use the git command line, you can try this open-source git learning game made by some people I know: https://ohmygit.org/ :slightly_smiling_face:

keunes:
Thanks for those git tips - will have a look!

PR #68, opened 2021-05-03
Update backup.md

strypey:
Added text clarifying that an OPML import adds subscriptions to the existing list, rather than replacing that list with the exported list.


_documentation/general/backup.md
```diff
@@ -17,4 +17,4 @@ Sync your subscriptions via [gpodder.net](https://gpodder.net/). This will save
...
 
 ## OPML
-OPML files allow to transfer your subscriptions to another podcast app. You will loose any podcast- or episode-specific data (like if a specific episode has been played) and all app settings. To export your file, go to `Settings` » `Storage` »  `Import/Export`. Be sure to click `Send…` to save the file in another folder on your phone or somewhere else, because the original file will be deleted when you uninstall AntennaPod.
+OPML files allow to transfer your subscriptions to another podcast app. You will loose any podcast- or episode-specific data (like if a specific episode has been played) and all app settings. To export your file, go to `Settings` » `Storage` »  `Import/Export`. Be sure to click `Send…` to save the file in another folder on your phone or somewhere else, because the original file will be deleted when you uninstall AntennaPod. When you import the file on another device, it will display a list with checkboxes and the podcasts you select will be added to your existing subscriptions.
```
ByteHamster, line 20: > it will display a list with checkboxes

Not really happy about that part, to be honest. Could you please also write that you can select which podcasts to import? Also, "it" sounds like it could mean any podcast app but we only know that about AntennaPod.

ByteHamster (commented):
Thanks for working on this! I left some feedback below :)

keunes (commented):
Thanks for your help! I had also left a comment, but apparently didn't 'submit' it :P Hoping this works.

strypey:
I've made some changes. Please let me know if the new language does the job. Do I need to submit a new pull request? I'm new to doing this stuff on GH ;)

ByteHamster:
> Do I need to submit a new pull request?

Nope, you did it just right :)

Looks good to me now.

PR #75, opened 2021-05-24
Add section title to blog posts

keunes:


ByteHamster:
Just tried it - works fine. I am not sure if we really need it, though. A big part of the screen is then filled with huge headlines.



How about using breadcrumbs like on the documentation page?

keunes:
Good point, that's actually neater. Let me see if I can figure that out :P

PR #83, opened 2021-07-01
Remove WiFi permission from privacy policy

ByteHamster:
We no longer use it since quite some time.


keunes:
Good point. For some reason the Netlify build is stuck in a Redirect loop: https://deploy-preview-83--antennapod-test.netlify.app/privacy I suppose you checked it builds correctly locally for you?

ByteHamster:
For GitHub pages, the last slash in the url matters, so we need the redirect to the version with slash. Apparently Netlify does some url cleanup automatically and removes the slash again.

PR #92, opened 2021-08-28
Troubleshooting

Matth7878:
Add a troubleshooting page / section.
If more troubleshooting as needed in future it will be possible to create sub section. For now I think having only 1 page is sufficient.

### DIFF
--- a/_documentation/troubleshooting.md
+++ b/_documentation/troubleshooting.md
@@ -1,0 +1,27 @@
+---
+title: Troubleshooting
+layout: doc
+level: "1"
+group: "troubleshooting"
+icon: "fas fa-bug"
+order: "70"
+---
+
+# Troubleshooting
+This page is for problems users can face.
+
+# Reporting a bug or an error in translation
+Refer to [this page](/documentation/general/bug-report)
+
+# Connectivity problems : podcast not updating or episodes not downloading
+Check if podcast link is working in your web browser. You can get the link used by AntennaPod in podcast screen, information icon and then tap on the podcast link to copy it.
+
+If the problem is with episode, try to download one with your web browser. Open link like previously described. Try to find download link to copy it and paste it in URL bar to trigger episode download. You can search for "enclosure url" or "mp3" or "type=" to help locating link.
+
+If it is not working with web browser then look on your device if there is not something interferring with network connection :
+- ad blockers like Blockada
+- firewall
+- VPN
+- DNS provider
+
+If it is with web browser but not with AntennaPod ask for help in [AntennaPod Forum](https://forum.antennapod.org/)


_documentation/troubleshooting.md
```diff
@@ -0,0 +1,27 @@
...
+title: Troubleshooting
+layout: doc
+level: "1"
+group: "troubleshooting"
+icon: "fas fa-bug"
+order: "70"
+---
+
+# Troubleshooting
+This page is for problems users can face.
```
ByteHamster, line 11: I would address users directly, something like "This page is for common problems you might face."

_documentation/troubleshooting.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Reporting a bug or an error in translation
+Refer to [this page](/documentation/general/bug-report)
+
+# Connectivity problems : podcast not updating or episodes not downloading
```
ByteHamster, line 16:
```suggestion
# Connectivity problems: podcast not updating or episodes not downloading
```

_documentation/troubleshooting.md
```diff
@@ -0,0 +1,27 @@
...
+Check if podcast link is working in your web browser. You can get the link used by AntennaPod in podcast screen, information icon and then tap on the podcast link to copy it.
+
+If the problem is with episode, try to download one with your web browser. Open link like previously described. Try to find download link to copy it and paste it in URL bar to trigger episode download. You can search for "enclosure url" or "mp3" or "type=" to help locating link.
+
+If it is not working with web browser then look on your device if there is not something interfering with network connection :
```
ByteHamster, line 21:
```suggestion
If it is not working with web browser then look on your device if there is not something interfering with network connection:
```

ByteHamster (commented):
Sorry for the long delay. I somehow forgot about the website changes. This needs a bit of re-structuring because the website changed but we can work on the text first.

@keunes, what do you think about the text?

keunes:
I'm so sorry! It seems I forgot about this PR as well - and in the meantime I had created https://antennapod.org/documentation/bugs-first-aid via #138. This PR addresses 2 points in the webpage:
* Reporting a bug or an error in translation --> Covered here: https://antennapod.org/documentation/bugs-first-aid/typo-translations
* Connectivity problems : podcast not updating or episodes not downloading --> Partly covered on https://antennapod.org/documentation/bugs-first-aid/episodes-missing. But 
  * We should take the trouble-shooting steps (basically all text below the section title) and add it to the current page.
  * I'm wondering if we should change the title of the current page ('Cannot download an episode') to something more generic ('Connectivity problems'). Or expand the existing title to include podcast refresh ('Can't download an episode/refresh a podcast'). Or maybe make a separate article for issues with podcast refresh, which could mention something about the exclamation mark that users often refer to. @ByteHamster, what do you think?

ByteHamster:
> We should take the trouble-shooting steps (basically all text below the section title) and add it to the current page.

Then the url of the page should probably be called "cannot download", not "episodes missing"

> something more generic ('Connectivity problems')

Seems a bit too generic to me. Users might have a hard time finding it.

> Or expand the existing title to include podcast refresh ('Can't download an episode/refresh a podcast').

:+1:

> Or maybe make a separate article for issues with podcast refresh, which could mention something about the exclamation mark that users often refer to.

Not sure if it is worth another page. I think pages that are mostly empty always look a bit weird. I would just mention it on the existing page.

PR #95, opened 2021-09-27
Update synchronization docs after 2.0.0 release

chk1:
The documentation still had some older text on the "how to synch" page, that is now obsolete since 2.0.0 has been released


ByteHamster:
Thanks!

PR #112, opened 2021-11-10
Improve site-readiness for translations

### LINKED ISSUE #111: Make site ready for integrated translation
**Short description**:
Make the website ready for integration in a (semi-automated) translation system.

**Location**:
Site-wide, except the blog (posts & section-specific interface).

**Why have this**:
Make the project and use of the app more accessible to audiences not comfortable with English.

**More info**:
List of components (Weblate) / resources (Pontoon):
* General page texts
* Contribute page texts
* Documentation page texts
* General interface texts

To do list:
- [ ] Split pages' MD files into 3 distinct containers w/in _i18n folder: Contribute, Documentation, General
- [ ] Transform localised navbar.yml (in _data) to single file with translation refs, include ref'ed items in generic yml file
- [ ] Make sure all UI texts in _includes are translatable (added as items to general _i18n/en.yml)
- [ ] Write CI script (GitHub action)

keunes:
Contributes to #111


_data/languages.yml
```diff
@@ -0,0 +1,6 @@
+- language:
+  code: en
```
ByteHamster, line 2: I don't know a lot of yaml but I think I read that the "language:" part should not be needed. You can directly add the dash before the code and remove the empty "language" key

```suggestion
- code: en
```

_documentation/getting-started/subscribe.md
```diff
@@ -1,35 +1,12 @@
...
 layout: doc
 level: "2"
 group: "getting-started"
 icon: "fas fa-plus"
 order: "10"
 ---
 
-# Subscribing to a podcast
-The first thing you want to do after downloading a podcast app is to subscribe to a podcast.
+# {% t documentation.categories.{{ page.group }}.{{ page.slug }} %}
```
ByteHamster, line 10: Hmm. I would actually change this to explicitly mention the key (`documentation.categories.getting-started.subscribe`) instead of building it based on other values

_includes/breadcrumbs.html
```diff
@@ -1,11 +1,11 @@
 <div id="breadcrumbs" class="pb-3">
     {% assign crumbs = page.url | remove:'/index.html' | split: '/' %}
-    <a href="/">Home</a>
+    <a href="/">{% t generic.home %}</a>
     {% for crumb in crumbs offset: 1 %}
       {% if forloop.last %}
-        / {{ page.title }}
+        / {% capture title %}{% t page.title %}{% endcapture %}{{ title }}
```
ByteHamster, line 6: That's a bit weird. You capture something just to print it directly afterwards. Wouldn't it be enough to just use `{% t page.title %}`?

_includes/documentation-card.html
```diff
@@ -0,0 +1,14 @@
+<div class="col-12 col-md-3 mb-3 mb-md-3">
+<a class="text-decoration-none" href="{{ include.item.url | replace:'.html','' }}">
+    <div class="card shadow-sm project-card">
+    <i class="{{ include.item.icon }} size-36 text-main mb-3"></i>
+    <h4 class="text-main weight-700">
+    {% if documentation.level == "1" %}
+        {% t documentation.categories.{{ include.item.group }}.title %}
+    {% elsif documentation.level == "2" %}
+        {% t documentation.categories.{{ include.item.group }}.{{ include.item.slug }} %}
+    {% endif %}
```
ByteHamster, line 10: Could be a lot less complex

```suggestion
    {% t {{ include.item.title }} %}
```

keunes: Ah, yes, apparently forgot to copy that from my test repo to this branch.

_documentation/getting-started/subscribe.md
keunes: Yes, `# {% t {{ page.title }} %}` seems to work :)

_documentation/automation.md
```diff
@@ -1,12 +1,12 @@
...
 group: "automation"
 icon: "fas fa-robot"
 order: "60"
 ---
 
-# Automation
+# {% t {{ page.title }} %}
 
-Are you an experienced user and want to automate tasks? Have a look at the pages below.
+{% t documentation.categories.{{ page.group }}.intro %}
```
ByteHamster, line 12: I would actually move this to markdown as well. Then we could add more detailed explanations when needed.

_includes/breadcrumbs.html
keunes: Didn't work before (I think document titles containing `{{ page.slug }}` caused this to fail), but now the document titles are settled with a full proper reference, `{% t page.title %}` seems to work fine indeed.

_documentation/automation.md
keunes: I think the introductions should be short, and without any markup. I want to style them a bit further (via CSS). Including them in the markdown files makes it more likely that the introductions will become messy.

_documentation/podcasters-hosters/list-podcast.md
```diff
@@ -1,18 +1,11 @@
 ---
-title: Getting your podcast listed
+title: documentation.categories.podcasters-hosters.add-on-antennapod
```
keunes, line 2: Incorrect ref; should be `list-podcast`

_i18n/en.yml
```diff
@@ -1,13 +1,47 @@
...
-  blog: "Blog"
   about: "About"
+  blog: "Blog"
   contribute: "Contribute"
-  license: "License"
   documentation: "Documentation"
+  download: "Download"
+  forum: "Forum"
+  license: "License"
   privacy: "Privacy Policy"
```
keunes, line 10: Double in the file

_i18n/en.yml
```diff
@@ -36,17 +70,29 @@ long_month:
...
-  download: "Download"
-  contribute: "Contribute"
-  view: "View All"
-  learn: "Learn More"
-  share: "Share"
-  error: "Error"
-  redirect: "Redirect"
-  subscribe: "Subscribe"
+language:
+  select: "Language"
```
keunes, line 74: Check: user where?

_data/languages.yml
```diff
@@ -0,0 +1,4 @@
+- code: en
+  name: English
+- code: de
+  name: Deutsch
```
ByteHamster, line 4: When done testing, this should be removed. The German translation is nowhere near ready

_i18n/en/general/license.md
```diff
@@ -1,4 +1,8 @@
-# Jekyll theme
+# AntennaPod
+
+AntennaPod is licensed under the [General Public License](https://github.com/AntennaPod/AntennaPod/blob/develop/LICENSE).
```
ByteHamster, line 3: Hmm, I wouldn't translate the license and other legal stuff (privacy policy etc)

contribute.html
```diff
@@ -1,19 +1,19 @@
...
 permalink: /contribute/
 ---
 
 <p>{% t contribute.introduction %}</p>
 
 <div class="container-fluid">
   <div class="row pb-4">
-    {% include contribute-card.html icon="fa-globe" url="https://www.transifex.com/antennapod/antennapod/" item="translate" %}
-    {% include contribute-card.html icon="fa-comment" url="https://forum.antennapod.org/" item="support" %}
+    {% include contribute-card.html icon="fa-globe" url="https://www.transifex.com/antennapod/antennapod/" item="translate" target="external"%}
```
ByteHamster, line 11: Space missing :)
```suggestion
    {% include contribute-card.html icon="fa-globe" url="https://www.transifex.com/antennapod/antennapod/" item="translate" target="external" %}
```

_data/navbar.yml
```diff
@@ -0,0 +1,11 @@
+# note: key must match page title key as used in _i18n/en.yml
+- key: about
```
ByteHamster, line 2: I think this should be `titles.about` instead. Otherwise, it is pretty hard to see that this is a translation key.

_i18n/en/documentation/queue/lock.md
```diff
@@ -0,0 +1,7 @@
+A pretty common problem experienced by users is that they suddenly can not manually reorder the queue.
+
+If you can't move episodes around in the queue, this can be caused by either of the following:
+
+- `Keep Sorted` is enabled. You can disable this by tapping the three dots `…` in the top bar and select `Sort`.
+- The queue is locked. You can unlock the queue by tapping on the `unlock` icon in the top bar:
+  ![Unlocking the queue](/assets/images/documentation/{{ site.lang }}/queue-unlock.png)
```
ByteHamster, line 7: I would specify the language explicitly here. Currently, the image is missing when selecting German because the file is not translated and therefore also does not have a localized image.

ByteHamster (commented):
A few things I noticed when skimming (not a detailed review yet)

_i18n/en/general/license.md
keunes: Yeah, I was hesitant about this one. I'll keep only the non-legal parts translatable. Our 'privacy policy' can be fully translated. I strongly doubt it has any legal value, and explains what data other parties have access to.

ByteHamster: Hmm, it should try to have legal value. At least Google requires adding a privacy policy and we link it.

_data/navbar.yml
keunes: Right now it's not a translations key, but used to create the translations key. But yeah, I can adjust so that the whole path is included here.

_i18n/en/documentation/queue/lock.md
keunes: > Currently, the image is missing when selecting German because the file is not translated and therefore also does not have a localized image.

This we should fix by adding the translated image ;) (But sure, we can mention the language explicitly here - we just need to make sure that the translators are using the right language code then, though (URLs of links are included in PO files by mdpo).)

_i18n/en/general/license.md
keunes: Sure, we try, but we don't have any legal lingo that might be lost in translation, which could have legal repercussions. We're simply explaining what AntennaPod does in layman's terms. Having such explanation in the user's language would be helpful to the user. If you want, we can add a sentence like 'in case of unclarity or inconsistencies between language versions, the English text will prevail' ([examples](https://www.lawinsider.com/clause/prevailing-language)).

_data/languages.yml
keunes: I actually would prefer the language menu `_includes/languages.html` use the `site.languages` array in the config file (instead of this `site.data.languages` array), and only take the language _names_ from elsewhere (e.g. this `languages.yml`). That'd make it less error-prone. Couldn't figure out a way (yet) though how to print value (languages name) from array y for each of the values in array x (site config).

_i18n/en.yml
keunes: `grep -r "language.select" *` didn't yield any result, so I think I had added it for something while trying things, but ended up not using it.

_data/languages.yml
keunes: Post created on the/a Jekyll forum, hoping for an answer :) https://talk.jekyllrb.com/t/display-data-from-array-x-in-for-loop-based-on-array-y/6677

keunes:
Hi @MichaelCurrin
Thanks again for your help with the simplification of the language menu over on the Jekyll forum. This is the PR I was working on (there's one commit that moves around a lot of files, which might be a bit distracting of the other changes). Much appreciated if you're willing to have a look, thanks a lot!

_includes/nav_languages.html
```diff
@@ -1,4 +1,4 @@
+{% for language in site.data.languages %}
 <a class="dropdown-item" onchange="location = this.value;"
-    href="{{ page.url | replace: '.html', '' | remove: 'index' }}">English</a>
-<!--<a class="dropdown-item" onchange="location = this.value;"
-    href="/de{{ page.url | replace: '.html', '' | remove: 'index' }}">Deutsch</a>-->
+    href="{% if language.code != 'en' %}/{{ language.code }}{% endif %}{{ page.url | replace: '.html', '' | remove: 'index' }}">{{ language.name }}</a>
```
MichaelCurrin, line 3: Readability improvement 
```suggestion
    href="{% if language.code != 'en' %}/{{ language.code }}{% endif %}{{ page.url | replace: '.html', '' | remove: 'index' }}">
        {{ language.name }}
    </a>
```

_data/languages.yml
MichaelCurrin: My suggestion on the forums was make languages.yml a hash object not a list.

```yaml
de: Deutch
en: English
```

---

Using

```yaml
languages: ["en", "de"]
```

Or the neater

```yaml
languages:
  - en
  - de
```

You can build the menu with

```liquid
{% for lang_code in site.languages %}
  {% assign lang_name = site.data.languages[lang_code]
  <a href="/{{ lang_code }}/">
    {{ lang_name }}
  </a>
{% endfor %}
```

MichaelCurrin: OR you have to do a filter from the list with longer and less clear code.


```liquid
{% for lang_code in site.languages %}
  {% assign lang_map = site.data.languages | where: 'code', lang_code | first %}
  <a href="/{{ lang_code }}/">
    {{ lang_map.name }}
  </a>
{% endfor %}
```


Reference for `where`:

https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/jekyll/liquid/all-filters.html#array-filters

keunes: Hi @MichaelCurrin, thanks again. In my forum post (reply to yours, which got automatically hidden), I wrote that I thought afterwards that languages should actually be sorted alphabetically - which can't be done (I think) when the forloop is based on `site.languages` (because the names aren't there). Instead I thought to keep `_data/languages.yml` an array with code and name, and have the following code to build the menu:
```
{% assign languages = site.data.languages | sort: 'name' %}
{% for language in languages %}
  {% if site.languages contains language.code %}
    <a class="dropdown-item" onchange="location = this.value;"
      href="{% if language.code != 'en' %}/{{ language.code }}{% endif %}{{ page.url | permalink: pretty }}">{{ language.name }}</a>
  {% endif %}
{% endfor %}
```
Do you think this makes sense?

MichaelCurrin: I would just manually keep the config languages sorted but yes if you want sort programmatically then this will work

I'd take out the permalink part as that doesn't do anything 

In your config you can set

```yaml
title: # ...
description: # ...
permalink: pretty 
```

And then page.url will be pretty already and the output files will be `/about/index.html/` rather than `/about.html`. So safer to do in another PR maybe.

Also if you use a subpath on gh pages then you need to add a prefix like this:

```
{{ page.url | relative_url }}
```

keunes: Ahh, I see now about the `permalink: pretty` I was already confused :) It seems that the output file already is `/about/index.html` because `about.md` has `permalink: /about/`. But I guess changing that to `permalink: pretty` throughout the site is less error-prone. Will do that in another PR :)

> Also if you use a subpath on gh pages then you need to add a prefix like this:

The lang menu links build fine locally without `| relative_url` - are you saying that it builds slightly differently on GitHub and requires this to be added?

I saved your dev cheatsheet btw, thanks, very handy!

MichaelCurrin: Oh yeah you can change on a per file basis
```
---
title: About
permalink: /about/
---
```

If this doesn't get what you want in the config:

```yaml
permalink: pretty
```

MichaelCurrin: In this case you don't strictly need `relative_url`.

Since your repo is `antennapod.github.io` and the site is `https://antennapod.github.io/`. (And in this case you get that redirecting to the `.org` URL)

If your repo actually had a name like `my-repo`, then the URL for the site root would be `https://antennapod.github.io/my-repo/` 

And assets would be on that too.
e.g.

```
https://antennapod.github.io/my-repo/assets/main.js
```

So in that case, `{{ '/assets/main.js }}` would _break_.

While `{{ '/assets/main.js | relative_url }}` would turn into a valid URL s `/my-repo/assets/main.js`

MichaelCurrin: Thanks for the cheatsheet bookmark, and feel free to add issues or PRs.

_i18n/en/general/license.md
ByteHamster: Yeah, I think adding such a sentence would be nice

_data/languages.yml
keunes: > If this doesn't get what you want in the config:
> `permalink: pretty`

It appears we already have `permalink: /:categories/:title`. At the same time we also have
```
collections:
  documentation:
    output: "true"
  blog:
    output: "true"
    permalink: /blog/:year/:month/:slug
```
I think the generic permalink setting is actually only relevant to the documentation, so I'll try moving it there, and making the generic permalink 'pretty'. But that's for a separate PR :)

Thanks again!

keunes:
Hi @ByteHamster

Did you still want to have a look? I'd like to merge it soon so this is out of the way :-)

ByteHamster:
- [ ] Please remove the German language from the menu. The translations there are just placeholders and not actual text. We don't want that to be indexed by Google.
- [ ] The search function does not show the titles. I think this can be fixed in search.json
  
- [ ] You renamed a link (auto_rewind to auto-rewind). There should be a redirect to the new URL.
- [ ] Please explicitly mention the language in included images. When translating the pages, we then need to pay attention on either creating the translated images or keeping the English images. Like it currently is, we always use the localized version, even if it does not exist. Alternatively, maybe it would be possible to write a function that checks if the image exists in the target language and if it doesn't, just falls back to English. That would be the easiest method - then we can even add `<!-- mdpo-disable -->`
  - queue/lock.md
  - general/beta.md
  - general/gpodder.md
  - download.md
- [ ] Translating the privacy policy is fine (with the "English prevails" sentence) but I still think that translating the license is not a good idea. I don't think there is much use for it and it could lead to problems. Licenses are worded carefully for a reason.

keunes:
Thanks again @ByteHamster for the feedback!

>  You renamed a link (auto_rewind to auto-rewind). There should be a redirect to the new URL.

It's a sub-page in the documentation. If for every update we ever make to the documentation requires a redirect, it might get a rather long list. I think such not-prominent pages do not deserve redirects.

> I still think that translating the license is not a good idea

The localised file has `<!-- mdpo-disable -->` above the license text, so it will not be translated.

I just discovered that there's a 'bug' in the menu rendering: the forum URL is prepended by the localised site URL. Once that's fixed as well, I'll commit the changes.

keunes:
Ok, changes committed :)

_documentation/general/beta.md
```diff
@@ -1,25 +1,11 @@
...
-
-So you want to have the latest and greatest? That's certainly possible! Simply visit the AntennaPod listing in the Play store and go to "Become a beta tester" at the bottom of the page. Or, if you're at your computer, you can go to [this url](https://play.google.com/apps/testing/de.danoeh.antennapod).
-
-<img src="/assets/images/documentation/project-beta.png" style="width:400px;" alt="Become a beta tester - confirmation window in Google Play">
-
-## F-Droid
-We do not currently offer beta versions on the F-Droid store.
-
-This is because F-Droid [does not](https://gitlab.com/fdroid/fdroidserver/-/issues/161) automatically recognise beta versions. Adding beta releases to F-Droid would require a manual intervention for each beta version. Also, if something important must be fixed, an update needs to be rolled out instantly. Unfortunately this cannot be guaranteed.
+{% capture content %}{% tf documentation/{{ page.group }}/{{ page.slug }}.md %}{% endcapture %}{{content | remove: "<!-- mdpo-disable -->" | remove: "<!-- mdpo-enable -->" | remove: "<!-- mdpo-disable-next-line -->" | remove: "<!-- mdpo-enable-next-line -->" }}
```
ByteHamster, line 11: Huh, why is this necessary here? (alternatively, why only here and not on all pages?)

_i18n/en/documentation/general/gpodder.md
```diff
@@ -0,0 +1,43 @@
...
+{% capture img-devices %}
+{% include image.html
+   alt="
+       <!-- mdpo-enable-next-line -->
+       devices
+
+       "
+   loc="/assets/images/documentation"
+   file="gpodder-devices.png"
+%}
```
ByteHamster, line 12: Hmm this is indeed quite a bit of text. Would it be possible to only include the URL, not the full image? Or does that add whitespaces? Then we do not need the mdpo enable/disable code.

Something like:

```
![devices]({% include localized-documentation-image file="gpodder-devices.png" %})
```

_documentation/general/beta.md
keunes: I just included the same code block stripping all mdpo tags from the pages where I used the image include. But having it on all pages is indeed more efficient. Where would you suggest to add it?

ByteHamster: Hmm, is it just to clean up the source code of the website or is it something that is visible to website users? If it is not visible to users, I would prefer using the shorter (easier to read/maintain) version.

keunes: As an HTML comment it's only visible if you inspect the site code. Still, it's a bit weird. But I can actually move it to the main layouts – much simpler to maintain :)
Also should update the `lang` while at it.

_i18n/en/documentation/general/gpodder.md
keunes: Downside is that there's a bigger chance that translators mess up, as they see the whole code rather than only the translatable text. Not sure whether risk of translators making a mistake is smaller/bigger than site code contributors. So I can change in line with your suggestion :)

keunes: It works but the URLs must be captured first, otherwise the images don't work due to spaces:
```
{% capture GP-url %}{% include image-url.liquid folder='badges' file='get-it-on-google-play.png' %}{% endcapture %}
[<img alt="Get it on Google Play" src="{{ GP-url | strip }}" height="80px"/>](https://play.google.com/store/apps/details?id=de.danoeh.antennapod)
![get on GP]({{ GP-url | strip }})
```

| |full image|url only|
|---|---|---|
|capture|only when code otherwise becomes hard to read and in lists|always|
|translators|deal only with text to translate|see & have to copy md tag (html tag on download page)|
|code complexity|more (particularly in md files)|less|

I'm inclined so use both includes: one for more full html image tags with class & size, one for url-only for md image tags.

ByteHamster: Hmm, having two different includes could be more confusing than helpful for contributors. I would say let's just keep it the way it is now :)

ByteHamster (approved):
Sorry for the late reply and so much back-and-forth with this PR. Looks good to me now, ready to merge :)

keunes:
Yeey. Thanks a lot for the feedback!

keunes:
Ah, actually: do you think we should/could squash into three commits?

* Everything before
* b2574f7 (where all the text gets moved to different files)
* Everything after

With the feedback I made quite a few subsequent changes on the same things - merging stuff would make it easier. But this one commit is just a lot of moving text around which would make reading history harder.

No idea how to squash certain commits only, though (only know on web you can squash all).

ByteHamster:
31b8eb9d7f0b0a8e60ba1ee7f8ee0bad120a77ba moves a lot of files, too. I squashed the last few commits that just said "review corrections", "final corrections" etc ;)

PR #126, opened 2022-01-02
Improve translations>master workflow

keunes:
Should probably have created an empty branch first instead of creating it with the workflow file. In any case, this PR updates the workflow file as the target branch name was wrong.

It is not very clean, but the Action to push updated translation files to main/master works on my branch. Please do give your comments on the whole workflow file, if you have any.

I used here cli git instead of the 'push subfolder' action as the folder structure is a bit complex. As far as I can see this only affects the commit author (would be Weblate when using 'push subfolder', now it is our very own 'bot').

And we should probably make translation-files a protected branch as well.

### DIFF
--- a/.github/workflows/translation-updates-to-main-repo.yml
+++ b/.github/workflows/translation-updates-to-main-repo.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Get main repository
         uses: actions/checkout@v2
         with:
-          ref: main
+          ref: master
       - name: Get translation repository
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


keunes:
Other actions to take:
* Set up a webhook so that Weblate gets informed of changes. Payload URL: https://hosted.weblate.org/hooks/github/
* Add user `weblate` ('Weblate (bot)') as collaborator

ByteHamster:
> Please do give your comments on the whole workflow file, if you have any.

GitHub only lets me comment on the changed parts of the file

How about using `antennapod-bot@users.noreply.github.com` as the email address? It then shows up as the bot account: https://github.com/antennapod-bot

keunes:
Ok, thanks, that's updated. Regarding the collaborator and webhook, can you apply them or give me the necessary rights?

ByteHamster:
Added weblate :heavy_check_mark:

PR #132, opened 2022-01-23
alt tags for website

comradekingu:
These currently show up as empty strings in Weblate https://hosted.weblate.org/zen/antennapod/website-short-strings/nb_NO/?offset=1&q=state%3A%3Ctranslated


keunes:
They don't indeed. That is because the images are just a crop of the main image and need to be replaced first. So I'll close this and I hope to prepare relevant images and according alt tags soon.

Thanks for the help!

PR #143, opened 2022-03-16
2.5 release blog

keunes:


ByteHamster:
Thanks! I do like the text but I'm not sure about the image. The license says it must not be cropped (I think the website does that, depending on the screen size) and must not be used in advertisements (which a Twitter post could be interpreted as).

I find Pixabay pretty nice for stock photos because of its permissive license (commercial allowed, no link required).

PR #160, opened 2022-06-05
Create documentation redirect

keunes:
Adresses https://github.com/AntennaPod/antennapod.github.io/pull/156#issuecomment-1145373537

### DIFF
--- a/documentation-redirect.md
+++ b/documentation-redirect.md
@@ -1,0 +1,6 @@
+---
+layout: redirect
+sitemap: false
+permalink: /documentation/
+redirect_to:  /documentation
+---


ByteHamster (approved):

keunes:
Hi @ByteHamster,

When building the site locally, I'm now getting an internal redirect circle if in the menu I click on 'Documentation' (`http://localhost:4000/documentation`) because it turns into `http://localhost:4000/documentation/` - which seems to be standard Jekyll behaviour - and which in turn (tnx redirect) sends you back to `http://localhost:4000/documentation`. And so the wheel turns. If I delete `documentation-redirect.md`, the issue is fixed. Now I'm wondering three things:
- do you see this happen, too?
- how come this eternal redirect is manifesting itself locally, but not on the live website?
- why did we have a 404 when opening the live site without the ending slash in the first place?

Any ideas?

ByteHamster:
I think the problem here is that GitHub Pages treats the URL with and without slash in the end as different pages, while most other web hosting providers (and also Jekyll Serve) don't care about the very last slash and still display the folder, even if the slash is not there. Some browsers might even add/remove slashes automatically...

PR #165, opened 2022-07-16
Create 10 year anniversary calendar event

keunes:


ByteHamster:
Why is this needed? Doesn't discourse allow downloading events? The downloaded file does not support repeated events but this 10 year anniversary should work fine from Discourse

PR #182, opened 2022-09-18
Added an initial release version + changelog

keunes:
This uses the jekyll-github-metadata gem and pulls the data from the GitHub API (it would need a periodic automated rebuild of the site in order to get the latest data; daily suggested). The content is then cached and does no 3rd party calls. Based on https://github.com/AntennaPod/antennapod.github.io/pull/163


ByteHamster:
- I find it a bit weird that the whole page moves when pressing that button.
- It is very prominent. More prominent than the download button. I would make it a lot less prominent.
- The line breaks seem to be missing

PR #184, opened 2022-10-03
Set up folder for redirects

keunes:
* Remove existing, old redirects — start with clean slate, 2 years after FAQ URL (https://github.com/AntennaPod/AntennaPod/issues/4740) was updated in the app this should be safe to do
* Add redirect for synchronisation documentation — for https://github.com/AntennaPod/AntennaPod/issues/6096
* Translate fall-back redirect link


ByteHamster:
/r/sync-documentation still seems a bit long. How about sync-help? Do we want r for redirect (technical term) or s for short?

keunes:
Happy to change it to `s` :-)

Will also change this slug. It was mainly for use in code to avoid problems when restructuring the support materials. But shorter is actually better as we can also use it in review comment replies :-)

PR #187, opened 2022-10-09
Add 2.7 release blog post

keunes:


_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
+---
+title: A new way to catch up on the latest episodes
+excerpt: Version 2.7, released today, introduces the Inbox!
+date: "2022-10-27 09:00:00"
+image: "2022/version2-7_new.jpg"
```
ByteHamster, line 5: Hmm, not sure about that picture. It feels too generic. For the previous versions, we had a "release banner" thing that we could also post on Twitter. (Template in branding repo)

_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
...
+author: keunes
+layout: blog
+---
+
+Whether you're subscribed to a lot of podcasts or just a handful, you want to know what's the next thing you can listen to/what to play next. Version 2.7 of AntennaPod introduces a better way to do just that: the Inbox!
```
ByteHamster, line 10:
```suggestion
Whether you're subscribed to a lot of podcasts or just a handful, you want to know what's the next thing to play. Version 2.7 of AntennaPod introduces a better way to do just that: the Inbox!
```

_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
...
+* Download episodes
+
+True podcast addicts might use the Inbox as a 'triage' screen, manually curating episodes and swiping the best ones straight into the queue. Those who follow just 2 or 3 podcasts might check this overview of new episodes when they're up to listening to something, and start playing an episode that looks interesting.
+
+If you're an existing user, you'll notice that your inbox will be full. That's because all episodes that had the 'new' label are now in your inbox. Whether you ignore the list, slowly work your way through it, or 'remove all from inbox' is up to you.
+
+# Not all that special
+Now you wonder - what's so special about an inbox? Don't we already know this concept since the nineties from e-mail?
+
+Truth be told: yes. But none of the other major podcast apps seem to have this. This functionality has been developed after a [request](https://forum.antennapod.org/t/introduce-new-screen-sytem-inbox/723) from one of our active contributors (the author of this post, as it happens), rather than by copying the other apps.
```
ByteHamster, line 28: Are you sure? I thought I had seen an inbox before somewhere

_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
...
+
+And we have to admint: even for AntennaPod it's not that revolutionary. Our long-time users probably already know the 'New' tab of the Episodes screen. The downside of the old Episodes screen was that the presence of tabs prevented the introduction of swipe actions. With that switched around managing new episodes is a whole lot easier.
+
+# What's next
+So far we've had positive feedback from our beta testers. (You can [join the beta testing](/documentation/general/beta) if you want.) But there already are ideas for further improvements. For example:
+* Allow to swipe episodes from the inbox directly into an 'ignored' status. These episodes then won't be auto-downloaded. Podcast junkies can build statistics about which podcasts they might want to un-suscribe from.
+* Introduce a 'Skip inbox' setting at podcast-level. Because the podcast is so great that you've already activated notifications for it, or don't need to triage them. Or because a podcast has only very few gems, and your friends will tell you about them.
```
ByteHamster, line 35: Do we want that on a per-podcast level? Activating notifications won't really help because the notification leads to the inbox screen. If you are only rarely interested, you can disable automatic updates (which also hides all episodes from the inbox).

_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
...
+* (Re-)Introduce a NEW label for podcasts that were released recently. So that you can see what's been released since you last opened AntennaPod.
```
ByteHamster, line 36: Hmm, not sure if I like that. Sounds like it will lead to a ton of unexpected behavior, for example when quickly opening the app to press "play" without actually looking at the episodes.

_blog/2022-10-08-2.7 release.md
```diff
@@ -0,0 +1,52 @@
...
+* Add a dating-app-like swipe experience. Because the list view doesn't provide much information, while a full-screen swipe experience allows to show (part of) the description text.
```
ByteHamster, line 37: Do we really want to announce this? Not sure if it will actually be implemented any time soon

_blog/2022-10-08-2.7 release.md
keunes: Yes, I explicitly chose not to go for the blue images any-more. (See 2.5 blog post.) They contain text which conflicts with (distracts from) the blog headings, and the whole page becomes boring blue.
I'll see if I can find a more appropriate image.

_blog/2022-10-08-2.7 release.md
keunes: > disable automatic updates (which also hides all episodes from the inbox)
Does it? If you do manual refresh episodes that are discovered then don't end up in the inbox?

_blog/2022-10-08-2.7 release.md
keunes: I've added a disclaimer.

_blog/2022-10-08-2.7 release.md
ByteHamster: I'm pretty sure they don't show up in the inbox, yes

_blog/2022-10-08-2.7 release.md
ByteHamster: Wouldn't it be cooler to surprise users with the feature instead of announcing it and then not implementing it for a long time?

Issue #183, opened 2022-09-25
Improve documentation on Synchronisation


keunes:
**Short description**: Describe that synchronisation must happen via a central point, a server. That these servers can be offered by big parties shared by many users (risk: slow if not funded). Or that you can self-host a server (downside: more effort). Describe that 'address of webserver' is the URL of the server on which you have an account.

**Location**: https://antennapod.org/documentation/general/synchronization

**Why have this**: 
> Works well enough, but synchronizing between devices with gpodder is completely broken, and documentation is non-existent. When I try to set it up, it gives me a blank text box and wants me to enter "Host name". What does that even mean?? No documentation.

https://play.google.com/apps/publish?account=8008695526664634386#ReviewDetailsPlace:p=de.danoeh.antennapod&reviewid=3a5cfe15-4392-406f-b882-6fef6e040bd8

**More info**:

Aliervo:
Love AntennaPod and would love to help out.  Mind if I take a pass at this?

keunes:
Hi @Aliervo,

That's nice, thanks! Feel free to do a PR or propose a text here as a comment.

loucasal:
This was completed in #191, so the ticket can be closed.

PR #191, opened 2022-10-20
Reword sync docs to avoid confusion over servers and hosts(#183)

Aliervo:
I reworded and re-ordered things to hopefully clear up some of the confusion as seen in #183. I also took a few in-app screenshots that I thought might help, but then decided that they would probably just end up bloating the page without actually adding much.  Happy to add them in if they are desirable.

### DIFF
--- a/_i18n/en/documentation/general/synchronization.md
+++ b/_i18n/en/documentation/general/synchronization.md
@@ -26,24 +26,25 @@
 
 <!-- mdpo-enable -->
 
-AntennaPod currently supports two synchronization options: via [gpodder.net](https://gpodder.net/) and the [gPodder Sync app for Nextcloud](https://apps.nextcloud.com/apps/gpoddersync).
+AntennaPod supports synchronization via [gpodder.net](https://gpodder.net/) and the [gPodder Sync app for Nextcloud](https://apps.nextcloud.com/apps/gpoddersync).
 
-On gpodder.net anyone can create an account relatively easily and the service is free to use. Unfortunately the service has large numbers of users, limited funding and a lack of volunteer contributions. This means the server often is overloaded, leading to errors in AntennaPod. If possible, we therefore recommend that you use an existing account on a Nextcloud instance or start self-hosting either gpodder.net or Nextcloud.
+gpodder.net is a free podcast synchronization web service that anyone can sign up for. Unfortunately, due to the popularity of the service and limited funding, the server is often overloaded, leading to errors in AntennaPod. We recommend that more technical users start self-hosting using either Nextcloud and gPodder Sync or their own gpodder service from the [github repo](https://github.com/gpodder).
 
-## gPodder Sync app for Nextcloud
-If you have a Nextcloud account, install the gPodder Sync app or ask your server admin to do so. Once that is installed, simply go to `Settings` » `Synchronization` in AntennaPod, choose the relevant provider and enter the server address.
+After creating an account or setting up a server, the following steps will get synchronization up and running:
+1. Set-up device(s) **gPodder ONLY**
+  - Log in to your account on gpodder.net (or on your self-hosted instance)
 
-## gpodder.net & your own gpodder server
-- Create an account on gpodder.net or log in if you have one already.
+  - Create a device under `Subscriptions` » `Devices` for each client that you use:
+    <!-- mdpo-disable-next-line -->
+    <br />{{ img-devices | strip }}
 
-- After you created the account, we suggest to create a device on gpodder.net/devices/ for each client that you use:
-  <!-- mdpo-disable-next-line -->
-  <br />{{ img-devices | strip }}
+  - When you have added the devices to your account, link them using the "Configure" button. This way, gpodder.net automatically keeps the subscriptions of the chosen devices synchronized.
+    <!-- mdpo-disable-next-line -->
+    <br />{{ img-synchronize | strip }}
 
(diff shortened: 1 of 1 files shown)


keunes:
Hi @Aliervo,

Many thanks for your PR. I just had a look.

I like the refinements in the first two paragraphs.
> We strongly recommend that more technical users start self-hosting using either Nextcloud and gPodder Sync or their own gpodder service from the [github repo](https://github.com/gpodder).

This feels unbalanced: we are referring people to the GitHub repo of gPodder.net, but not of Nextcloud. We should redirect to both or to none. If we do want to add links, I would link to the installation instructions of each service.
* gpodder.net: https://gpoddernet.readthedocs.io/en/latest/dev/installation.html
* Nextcloud: https://nextcloud.com/yourdata/ or https://nextcloud.com/install/#instructions-server

Then, I'm not sure about the reorganisation of the instructions themselves. I understand that it is now a single flow, with 'conditional' sections (only if gPodder.net...). However, as a consequence, if you're using Nextcloud, you have to skim a lot of text/page just to read 2 lines:
Screenshot 2022-10-26 at 09-07-00 How to synchronize between devices – AntennaPod

I would rather revert those changes, and keep the current separation between gPodder and Nextcloud. It is much faster to go through if you're a Nextcloud user and for all users takes away the reflection about each condition.

Lastly, I think it would be great if we could still address the notes in the issue you referred to:
* _Describe that synchronisation must happen via a central point, a server. That these servers can be offered by big parties shared by many users (risk: slow if not funded). Or that you can self-host a server (downside: more effort)._ --> This I guess best goes in the introduction, or at least the first point (need of a central server) - the rest is already kinda covered in the sentences about why gPodder often leads to errors.
* _Describe that 'address of webserver' is the URL of the server on which you have an account._ --> This you already covered in brackets, but I think we can make it more prominent. For example by splitting it off into a separate sentence.

Would you be OK with following up on these comments?

Aliervo:
Thanks for the feedback, I'll make a few edits!

> This feels unbalanced: we are referring people to the GitHub repo of gPodder.net, but not of Nextcloud. We should redirect to both or to none. If we do want to add links, I would link to the installation instructions of each service.
>
> 
> * gpodder.net: https://gpoddernet.readthedocs.io/en/latest/dev/installation.html
> * Nextcloud: https://nextcloud.com/yourdata/ or https://nextcloud.com/install/#instructions-server

I feel like I meant to do this and just forgot about poor Nextcloud. Also, I got a little confused with the gPodder installation instructions because they seem relatively straightforward there, but on then I looked at the GitHub and noticed some repos related to mygpo, namely "feedservice" and "auth", and I wasn't sure if all three services or just the main repo would be needed to get sync working. I'll try to get it running to figure out what exactly to link.

> Then, I'm not sure about the reorganisation of the instructions themselves. I understand that it is now a single flow, with 'conditional' sections (only if gPodder.net...). However, as a consequence, if you're using Nextcloud, you have to skim a lot of text/page just to read 2 lines: Screenshot 2022-10-26 at 09-07-00 How to synchronize between devices – AntennaPod
> 
> I would rather revert those changes, and keep the current separation between gPodder and Nextcloud. It is much faster to go through if you're a Nextcloud user and for all users takes away the reflection about each condition.

I also noticed this once I got it all together. It almost made me want to change the first part to "Just use Nextcloud, it's easier." :rofl:

Gonna try to figure out mygpo installation, expect some edits soon-ish. :+1:

keunes:
> Just use Nextcloud, it's easier.


Hehehe. We can indicate that indirectly, by putting it first & short :)

Aliervo:
@keunes, I think this might be the ticket. I re-wrote the intro to be pretty explicit about servers and hostnames and added more links.  I didn't quite revert the changes to the actual instructions, but instead broke the gPodder specific "Set-up steps" to their own section below the "main steps" so we can maintain one flow, while preventing the Nextcloud folks from having to read most of the page to find what is relevant to them.

I was not able to set up my own ``mygpo`` server because all the instructions and even the scripts it uses are Ubuntu/Debian based; I am currently on Arch and didn't want to waste all night figuring out what the Arch equivalents of all the libs are, but I linked the page anyway.  If we get complaints, we might just have to write our own guide or specifically request Nextcloud. 

My only concern with Nextcloud link/instructions is that they put "Sign up with a provider" higher on the page than the actual self-hosting options, which I don't believe allows installing custom apps. My hope is that people tech-savvy enough to self-host will be smart enough to choose the correct option for their situation.

Let me know if there is anything I missed!

keunes:
Hi @Aliervo,

Sorry for not getting back to you earlier. I had another look today and I still think that having two sections with cross-references between them is not user friendly. So I took the liberty to change the approach again, even though it leads to somewhat longer texts. I have also adjusted the text based on https://github.com/AntennaPod/AntennaPod/pull/6213.

> My only concern with Nextcloud link/instructions is that they put "Sign up with a provider" higher on the page than the actual self-hosting options, which I don't believe allows installing custom apps.

The page currently doesn't describe this option. I was hesitating to add it, but given the limitation you describe with app installations and the fact that we have no idea if/how this works, I decided to leave it out for now. Someone on the forum did indicate they were looking into this option: https://forum.antennapod.org/t/some-help-with-sync-please/2524. Maybe if we get some results/information from that, we can add this option as a second bullet point.

If you are OK with it as well, I am happy to merge this as soon as the next AntennaPod version gets released when the PR I linked gets available to everyone (hopefully early next year).

PR #192, opened 2022-10-20
Add events to website

keunes:
If ever we change the recurrence, we should do something with `RECURRENCE-ID` (https://icalevents.com/4437-correct-handling-of-uid-recurrence-id-sequence/). Not sure if a new VEVENT block would be required or not.
* [ ] Check for a GitHub action to check ICS validity, maybe system that checks if it doens't exist yet.

Key resources that helped me understand ICS & helped me make this:
* https://www.webdavsystem.com/server/creating_caldav_carddav/calendar_ics_file_structure
* https://stackoverflow.com/a/37348467
* https://icalendar.org/RFC-Specifications/iCalendar-RFC-5545/
* https://talk.jekyllrb.com/t/how-to-fold-wrap-the-content-after-number-of-characters/7312/6?u=keunes
* https://github.com/vinorodrigues/bootstrap-dark
* https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/jekyll/liquid/time-handling.html
* http://www.strfti.me/
* https://www.shopify.com/partners/blog/liquid-date-format


_config.yml
```diff
@@ -34,6 +34,21 @@ collections:
   blog:
     output: "true"
     permalink: /blog/:year/:month/:slug
+  events:
+    output: "true"
+    permalink: /events/:year/:month/:title
+
+defaults:
+  - scope:
+      path: ""
```
ByteHamster, line 43: This looks like it would be possible to use different folders for blog posts and events. If that's the case, I think that would be nicer :)

_events/community-meeting.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Community meeting
+summary: Our monthly meeting to discuss anything AntennaPod. Join us!
+image:
+permalink: /events/:slug
+date: "2022-04-09 18:00:00 +0200" # yyyy-mm-dd hh:mm:ss _hhmm Where _ is either + or - to indicate the timezone in which the event is noted
+ICS: 0361F458-FD97-46CD-A014-076C1995439E # Name of file added in assets/events folder
```
ByteHamster, line 7: Instead of having one ICS file per event that one needs to add to their calendar, how about having one single, dynamic ICS that contains all (old and new) events? We could even automate the creation of that file, based on the metadata given above. Then one needs to add it to their calendar only once, and automatically gets new events showing up

_data/navbar.yml
```diff
@@ -9,3 +9,5 @@
   url: /documentation
 - key: titles.forum
   url: https://forum.antennapod.org/
+- key: titles.events
+  url: /events
```
ByteHamster, line 13: I think it would fit better before or after "contribute". Then the last item stays the "external" link. This also ensures that the "contribute"-type links are together, while the "user-support"-type links are in the end.

keunes:
@ByteHamster The Netlify deploy failed because our deploy setting was to use Ubuntu 16.04 and it's depricated. I've changed it to 20.04 and the deploy succeeded (messages here are just not updated).
https://deploy-preview-192--antennapod-test.netlify.app/

keunes:
What I don't get:

Putting https://deploy-preview-192--antennapod-test.netlify.app/events.ics in this validator: https://icalendar.org/validator.html gives an error:
> Lines not delimited by CRLF sequence [near line # 1](javascript:void(0);)
> Reference: [RFC 5545 3.1. Content Lines](http://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html)

However, when copy-pasting the file contents, the validator doesn't throw an error.

keunes:
Trying to subscribe to https://deploy-preview-192--antennapod-test.netlify.app/events.ics as an external calendar throws an error :(

ByteHamster:
I don't know much about ics, but should this semicolon maybe be a colon?

```
RDATE;VALUE=PERIOD;TZID=Europe/Brussels:20221022T160000/20221022T170000
                  ^
```

ByteHamster:
> Lines not delimited by CRLF sequence

Try changing the line endings of the file in your editor. Apparently ics files need Windows line endings.

keunes:
> I don't know much about ics, but should this semicolon maybe be a colon?

Not according to [the standard](https://www.kanzaki.com/docs/ical/rdate.html#descr). And I saw this example here: https://github.com/ical4j/ical4j/issues/549

> Try changing the line endings of the file in your editor. Apparently ics files need Windows line endings.

Ahh, thanks! Will give that a try :-)

ByteHamster:
> Not according to [the standard](https://www.kanzaki.com/docs/ical/rdate.html#descr).

In the examples in the bottom, they do use a colon.

keunes:
> In the examples in the bottom, they do use a colon.

The official example has only one attribute each time. In case of multiple attributes, only at the end the colon is used:
`ATTENDEE;ROLE=REQ-PARTICIPANT;PARTSTAT=TENTATIVE;CN=Henry Cabot:MAILTO:hcabot@host2.com` ([example](https://www.kanzaki.com/docs/ical/attendee.html))

_events/0_A meeting template.md
```diff
@@ -0,0 +1,49 @@
...
+# Recommended. Human-readable description of the location (e.g. address) for ICS. If not provided, link to meeting-room or to OSM will be used.
+more-information: 
+# Optional. Must be a URL.
+meeting-room: 
+# Optional. Must be a URL. Link displayed on event page if "location: online".
+blog-post: 
+# Optional. Must be the path of the blog post, starting with the year (e.g. /2022/10/translation-coordinator).
+---
+
+# Event description goes here. Displayed on the event detail page, and used for the body of the ICS calendar event.
```
ByteHamster, line 49: How is a heading handled in the event description? Isn't this somehow doubled?

_events/0_A meeting template.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Awesome get-together on topic X 😎
+# Required. Displayed both on the event detail page and in the ICS calendar event.
+summary: 
+# Required. Displayed on the Events overview page.
+image: 
+# Optional. Displayed on Events overview page and the event detail page.
+permalink: 
+# Optional. "/events/:slug" recommended for recurring events, to avoid an an event getting a date-based URL.
```
ByteHamster, line 9: Does renaming then change the slug? Maybe the recommended value should be to explicitly write something?

_i18n/en/general/event-sidebar.md
```diff
@@ -0,0 +1,9 @@
+## Meet us!
+
+Are you interested in meeting the people behind AntennaPod? You'll find all the opportunities here on this page!
+
+Do you know of an event in your neighbourhood and would love to meet up with AntennaPod contributors and users? Tell about the event on our forum to organise a get-together!
+
+[Propose a meet-up](https://forum.antennapod.org/)
+
+<a href="{{ site.baseurl }}/events.ics" target="_blank" data-toggle="modal" data-target="#icsModal">Add our events to your calendar</a>
```
ByteHamster, line 9: Exclude tag from translations?

_includes/ICS-modal.html
```diff
@@ -0,0 +1,33 @@
+<script type="text/javascript" src="/assets/js/copy-url.js" />
```
ByteHamster, line 1: Could it happen that the script is included twice this way? That could lead to problems. I would rather include it in the main header and nowhere else.

_i18n/en/general/events-add-calendar-modal.md
```diff
@@ -0,0 +1,5 @@
+Always want to have the latest communinity call dates and times in your calendar? Then copy the URL and add it as a remote calendar in your calendar app.
+
+(For the moment we don't have the option add individual events to your calendar.)
```
ByteHamster, line 3: I would remove that. Explicitly specifying what we don't support is just overhead for readers.

_events/2022-07-22-10 year anniversary.md
```diff
@@ -0,0 +1,17 @@
+---
+title: 10 Year anniversary 🎉
```
ByteHamster, line 2: Set permalink to `events/10YearAnniversary` to keep links working?

keunes: Meh, I would rather keep the file structure clean - happy to accept a one off article or toot with a broken link. That URL was only to download the ICS anyway.

_includes/ICS-modal.html
keunes: Quite unlikely but possible (e.g. if we add the ICS modal to a page with sharing options, or add sharing options to the events page). I included it here to avoid the js being loaded on every page, but I guess it's small enough not to be an issue.

_events/0_A meeting template.md
keunes: Renaming the file does not change the URL - it is based on the title: `/events/:year/:month/:title`. I think that's fine. The template already is quite complex, so I want to keep as few lines required or recommended as possible.

_events/0_A meeting template.md
keunes: HTML tags are stripped from the ICS file, and in the event page the MD should be rendered as expected. I put the # there to indicate a comment but as here it's part of the body I guess that's misleading. Removing the #

_events/2022-07-22-10 year anniversary.md
ByteHamster: I find it a bit sad to possibly break links, but okay

_events/0_A meeting template.md
ByteHamster: I didn't mean adding additional recommendations. I was just thinking about changing the recommendation that is already there

PR #201, opened 2022-11-26
Fix page redirects

ByteHamster:
This mostly reverts 61bccc5e40d87df72d5af49284d1afdf13308aa8


keunes:
Why do we need the redirect with trailing slash again? I forgot - was it included in some version of the app? I don't really like the aesthetic of the trailing slash, and it's also inconsistent with the other pages/sections (e.g. the blog).

I've posted also on https://talk.jekyllrb.com/t/no-trailing-slash-just-fine-on-github-pages-but-eternal-redirect-locally/7851

ByteHamster:
The problem is the difference in how GitHub handles hosting. If it's generated as a folder (with slash), users can visit both the page with and without slash. If it is generated as a file, GitHub only serves the version without slash. Adding a slash then shows an error message. That's consistent with what other web servers do.

PR #213, opened 2023-01-17
Change path symbol in Update privacy.md

femmdi:
Changing symbol in (Settings → Backup & Reset → Back up my data) from "→" to "»" as suggested by @loucasal on weblate: https://hosted.weblate.org/translate/antennapod/website-general/en/?q=has%3Acomment&offset=3


ByteHamster:
I wonder if, in the long run, we should switch all from `»` to `→`. The reason is that `»` conflicts with French quotation marks, which caused issues in the past (people "translating" `»` to `"`, probably by blindly accepting the machine translation). What do you think @keunes?

PR #215, opened 2023-01-18
Use RDATE without PERIOD

ByteHamster:
I guess it is quite unusual to not only change the time, but also the duration of a meeting. Let's just use RDATE without PERIOD, so it is supported on more calendar applications.


ByteHamster:
Hmm, apparently this does not help :/ Still getting an error message in ICSx5

PR #220, opened 2023-02-26
Brand update

keunes:
TODO:
* [x] edit brand colours in CSS
* [ ] Check share image: https://mastodon.social/@keunes/109933138422181631
* [ ] Write announcement & thank blog post
* [x] Fix navbar mobile layout:


ByteHamster:
I think something is wrong with the cover image. It has transparency.

keunes:
> I think something is wrong with the cover image. It has transparency.

Hmm. I didn't realise. But actually, it's not so bad; makes the default images (for blogs/events) fit slightly better with the theme, as they are not too dark/bright.

Is this transparency causing an issue somewhere?

ByteHamster:
I don't know. This feels more like a bug than a feature. The other blog images are not transparent either.

PR #222, opened 2023-04-16
Add dependabot & update GH Actions dependencies

### LINKED ISSUE #199: Look into actions/checkout warning for Translations action
**Short description**: Look into actions/checkout warning

**URL**: 

**More info**: 
Used in GH action to pull translations from `translations` to `main` branch.
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2

Rotzbua:
Currently there is no package manager or similar for GH Actions to manage the dependencies.
Dependabot offers to monitor the workflow scripts and propose new versions as PR.
Since GH Actions rarely change and no resources should be wasted, a repetition of one month is sufficient.

I would recommend to add the dependabot.yml to every repository with GH action script.

## Reference

fixes #199

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 2
+    assignees:
+      - "keunes"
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
```
keunes, line 11:
```suggestion
  - package-ecosystem: "github-actions" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "monthly"
    open-pull-requests-limit: 2
    assignees:
      - "keunes"
```

keunes (commented):
Thanks @Rotzbua for thinking about us. I suggest assigning PRs automatically and having maximum 2 open PRs.

@ByteHamster any thoughts? (As we shouldn't expect too many updates I think it's fine. Can be a nice reminder for the side work that we don't look into actively too often.)

PR #223, opened 2023-04-16
Fix validator.w3.org warnings and erros

Rotzbua:
* Info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
* Warning: The type attribute is unnecessary for JavaScript resources.
* Error: An img element must have an alt attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
* Error: Bad value for attribute width on element img: The empty string is not a valid non-negative integer.
* Error: Bad value 260px for attribute width on element img: Expected a digit but saw p instead.
* Warning: The button role is unnecessary for element button.

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -3,7 +3,7 @@
         <div class="navbar navbar-expand-xl justify-content-between">
             <div>
                 <a href="/{% if site.lang != 'en' %}{{ site.lang }}{% endif %}">
-                    <img class="smallermobile" src="/assets/logos/logo.png" id="mainlogo" />
+                    <img class="smallermobile" src="/assets/logos/logo.png" id="mainlogo" alt="{% t generic.accessibility.logo %}"/>
                 </a>
             </div>
             <div class="collapse navbar-collapse flex-grow-0 weight-500">
@@ -16,7 +16,7 @@
                     <div class="nav-item ml-0 ml-sm-1 text-center">
                         <li class="nav-item dropdown mb-0">
                             <button class="nav-link dropdown-toggle" id="navbarDropdown" aria-label="{% t generic.accessibility.lang-menu %}"
-                                role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                                data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                                 <i class="fa fa-globe size-16 color-gray-900"></i><span
                                     class="ml-1 size-16 color-gray-900 weight-600">{{ site.lang | upcase }}</span>
                             </button>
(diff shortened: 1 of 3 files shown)


_includes/nav.html
```diff
@@ -3,7 +3,7 @@
         <div class="navbar navbar-expand-xl justify-content-between">
             <div>
                 <a href="/{% if site.lang != 'en' %}{{ site.lang }}{% endif %}">
-                    <img class="smallermobile" src="/assets/logos/logo.png" id="mainlogo" />
+                    <img class="smallermobile" src="/assets/logos/logo.png" id="mainlogo" alt="{% t generic.accessibility.logo %}"/>
```
keunes, line 6: Hmm. We already use these strings to define an area-label for the menu as a whole. Is the alt on the image then necessary still? Or will users with screen readers now hear the same info twice? 🤔

Rotzbua: I am not an expert for accessibility. I do not know if screen reader user will hear it twice. But the html does not not have an area-label. The alt text is also used for users which have images disabled.

keunes: > But the html does not not have an area-label.

Right, sorry. That's for the language menu, line 18:
`<button class="nav-link dropdown-toggle" id="navbarDropdown" aria-label="{% t generic.accessibility.lang-menu %}"`

Then, for the logo: if we don't have the text included anywhere in the header yet, which indeed seems to be the case, then I wonder why we have the string in the localisation file. I think we must have used it somewhere…

Also, you updated the header but I believe, if we add the alt text to the logo, the footer should be updated also.

Rotzbua: > That's for the language menu, line 18:
<button class="nav-link dropdown-toggle" id="navbarDropdown" aria-label="{% t generic.accessibility.lang-menu %}"

You are right. That label is wrong. Have to update.

> Then, for the logo: if we don't have the text included anywhere in the header yet, which indeed seems to be the case, then I wonder why we have the string in the localisation file. I think we must have used it somewhere…

The logo alt text is from https://github.com/AntennaPod/antennapod.github.io/blob/e7a8da891122244b7c9cdcd6fdf0c9de9c656e50/_includes/nav.html#L51-L52

_includes/nav.html
```diff
@@ -40,16 +40,16 @@
                         </div>
                     </div>
                 </div>
-                <button class="navbar-toggler" onclick="openNav()"><img
-                        src="{{ site.baseurl_root }}/assets/images/icons/burger.svg" /></button>
+                <button class="navbar-toggler" onclick="openNav()">
+                    <img src="{{ site.baseurl_root }}/assets/images/icons/burger.svg" alt="{% t generic.accessibility.lang-menu %}"/>
```
Rotzbua, line 44: TODO add translation or later PR: generic.accessibility.menu

PR #227, opened 2023-04-25
Fix validator.w3.org warnings and erros

Rotzbua:
* Error: Bad value 80px for attribute width on element img: Expected a digit but saw p instead.
* Error: Bad value auto for attribute height on element img: Expected a digit but saw a instead.
* Error: Bad value for attribute width on element img: The empty string is not a valid non-negative integer.
* Error: Bad value 260px for attribute width on element img: Expected a digit but saw p instead.
* Warning: The button role is unnecessary for element button.

## Reference

Part from original pr #223


keunes (commented):
For consistency I'd just keep the spaces, but otherwise I guess it can be merged. Thanks for working on this :-)

keunes:
Probably better in a separate PR, but if you see a way to 'merge' the menu desktop- and mobile versions we can get rid of the 'Duplicate ID' error. (I wouldn't really feel comfortable attempting that - don't know if you would.)

Rotzbua:
> For consistency I'd just keep the spaces, but otherwise I guess it can be merged. Thanks for working on this :-)

Uncommon but I try to follow the style next time 👍

keunes:
>> For consistency I'd just keep the spaces, but otherwise I guess it can be merged. Thanks for working on this :-)

> Uncommon

I'm not a web development expert. If it's uncomment, then I'm happy also to move to the non-space standard. As long as we're consistent :-)

ByteHamster (approved):

Issue #203, opened 2022-11-29
Upgrade to Bootstrap 5.3+


keunes:
**Short description**: Upgrade Bootstrap to version 5.3 or later, when it is released.

**Location**: 

**Why have this**: It 'natively' supports Dark mode 🎉 This will decrease the need for custom CSS. And it will undoubtedly bring other improvements as well.

**More info**:
 https://github.com/twbs/bootstrap/pull/35857#issuecomment-1330196729

ByteHamster:
Also, building now gives these warning messages because of the outdated bootstrap:

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($spacer, 2) or calc($spacer / 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

    ╷
717 │ $navbar-padding-y:                  $spacer / 2 !default;
    │                                     ^^^^^^^^^^^
    ╵
    /mnt/DATEN/Entwicklung/htdocs/antennapod.github.io/_sass/bootstrap/_variables.scss 717:37  @import
    bootstrap/bootstrap.scss 9:9                                                               @import
    /mnt/DATEN/Entwicklung/htdocs/antennapod.github.io/assets/css/main.scss 4:9                root stylesheet
```

Rotzbua:
Another advantage is the removal of jQuery dependency. For custom code this can be done upfront.
Reference: https://getbootstrap.com/docs/5.3/migration/#dependencies

keunes:
Hi @Rotzbua,
With #203 done, are we ready to start working on this? (If yes, I'm wondering how to approach this. Maybe we should see if we can drop more custom CSS? Also, would you be up for a call to discuss the site/this upgrade in person?)

keunes:
Hi @Rotzbua, kind reminder. Are you interested in working in this? I would love to work on this :-) (if no time or interest, please don't hesitate to mention that also)

Rotzbua:
@keunes I will have a look around Christmas and new year

PR #230, opened 2023-04-30
Upgrade to Bootstrap v4.6.2

Rotzbua:
## Changes

* Fixes warning: Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.
* Moved bootstrap customization to `_variables.scss`
* Update bootstrap to latest v4 version
* Update jQuery to latest version

## Reference
Preparation for migration to v5 #203


ByteHamster:
Can't we just switch directly to 5.0 instead of taking two individual steps?

Rotzbua:
I am not a fan of a mayor version migration with skipping subversions. Because then it is difficult to find the source of a problem.
Migration to bootstrap 5 does require some refactoring. This pr is just library update without any code change but fixes the sass warnings.

_sass/bootstrap/_variables.scss
```diff
@@ -212,7 +212,7 @@ $container-max-widths: (
   sm: 540px,
   md: 720px,
   lg: 960px,
-  xl: 1360px
+  xl: 1140px
```
keunes, line 215: Given the smaller max-width for containers we should update the widths of the event cards. I had tweaked them to look OK-ish, now they're too small: https://deploy-preview-230--antennapod-test.netlify.app/events/

Similarly, we might want to limit the number of ['Contribute' cards](https://deploy-preview-230--antennapod-test.netlify.app/contribute/) to 3 (rather than 4).

keunes (commented):
Thanks! I think we need to make some minor changes. But otherwise: let's merge :-)

_sass/bootstrap/_variables.scss
Rotzbua: It seems that bootstrap 4.5.2 was modified. I restored the vanilla bootstrap 4.5.2 and moved the customization to `_variables.scss`.
Now the result looks like before, except the corner radius.

keunes: Actually I prefer the vanilla bootstrap - it's less wide which makes it easier to read. So I would keep that and adjust the number of event and contribute cards accordingly.

Rotzbua: I would favor leaving this PR as a mostly technical upgrade and consider the visual changes when migrating to bootstrap v5.

keunes:
Thanks again :-)

PR #232, opened 2023-05-06
Explain why F-Droid takes a while to build

ByteHamster:
Maybe this can prevent at least some of the questions we get for every single release...


_i18n/en/documentation/general/f-droid.md
```diff
@@ -1,9 +1,10 @@
-After every new release of AntennaPod, we get asked when it will be available on F-Droid. Releases on F-Droid take a few days, so please be patient.
+After a new version of AntennaPod gets released, we often are asked when it will be available on F-Droid. We can't give an exact answer — it usually takes a few days.
```
ByteHamster, line 1: I would still keep the "please be patient". If it's a bit more explicit, maybe this stops people from asking again and again.

PR #234, opened 2023-05-07
Upgrade to font fontawesome v6

Rotzbua:
- Update font fontawesome to 5.15.4
- Upgrade font fontawesome to 6.4.0
- Migrate fontawesome v6 classes and names


keunes:
I was thinking earlier (before you started working on this) that it's a bit odd that we have a `css` folder as well as a `sass` one, and that FontAwesome, while technically a font, doesn't really belong in a font folder. How would you feel about dropping the `css` folder and moving all FontAwesome files in a single `FontAwesome` folder?

Rotzbua:
I think the `css` folder is fine, because it is the public folder.
I would move the `all.css` to the folder `fonts/fontawesome`, so font files and font css definitions in the same place.

PR #236, opened 2023-05-08
Add Media kit & Branding page

keunes:
And publish the first press release.

Note: This PR doesn't (yet) update the GH actions required for translations. A separate PR will be required.

PR #237, opened 2023-05-08
Fix Download page image formatting to actually link to F-Droid and Google Play

nathan-contino:
Hi!

I noticed that your download page is currently kinda broken. The URLs exist, but there's some broken markdown showing on the page itself. But you don't have to take my word for it: https://antennapod.org/download/

I took the liberty of fixing the issue (on all languages) by passing a URL argument to the image display code. Existing images that don't pass a URL will still display a modal exactly as before. But images that *do* pass a URL will now load that URL on click (instead of the modal).

I'm not sure if there's any hoops I have to jump through to get this contribution accepted, but let me know if I need to sign any agreements or adjust my commits to get this change in. Thanks!

### DIFF
--- a/_i18n/da/general/download.md
+++ b/_i18n/da/general/download.md
@@ -1,26 +1,37 @@
-{% capture img-GP %} {% include image.html alt="
+<!-- mdpo-disable -->
 
+{% capture img-GP %}
+{% include image.html
+   alt="
+       <!-- mdpo-enable-next-line -->
        Få den på Google Play
 
        "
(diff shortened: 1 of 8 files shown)


keunes:
Hi @nathan-contino,

Thanks a lot for spotting this and raising a PR! I had some other pending changes in #236 which would be incompatible with your PR, so I created another PR (with those pending changes and the correction needed) which will close your PR.

Best

nathan-contino:
Thanks! Good luck with the big 3.0 release tomorrow.

PR #242, opened 2023-05-10
Change file name to lower-case

Rotzbua:


keunes:
Any particular technical reason, or just for consistency?

Rotzbua:
Both reasions. Linux is case sensitive and I wasted time with case-typos containing the chars: lLIO0 😉

PR #244, opened 2023-05-14
 Add `.editorconfig` for code style consistency 

Rotzbua:


ByteHamster:
Before reformatting files, I think we should have a CI check that enforces the formatting. Otherwise the files will soon be inconsistent again

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
```
ByteHamster, line 8: @keunes What indent do we want to use? Most files use 2 spaces, some files use 4 spaces. We should make this a deliberate decision.

keunes: Personal preference? 2.

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{js,html,md}]
+max_line_length = 500
```
keunes, line 14: Is there any benefit to this? I would be inclined not to have a line limit, so that each text block is on a single line in all source and translated files.

Rotzbua:
> Before reformatting files[..]

I removed the commit which formats some files from this PR to avoid conflict with other PRs. Formatting the files can be done by you so it does not require extra review.

>  I think we should have a CI check that enforces the formatting.

I never used a lint tool for editorconfig. Until now, I considered editorconfig to be an automatic text editor configuration rather than a strict style rule.
But I am open to new useful tools if they exist?

.editorconfig
Rotzbua: My intention was to avoid the hard wrap configured by some editors. It seems that Intellij works with `max_line_length = off`.

Issue #247, opened 2023-05-23
Check links are working for each commit/PR


keunes:
**Short description**: Set up a CI action to check links are working for each commit/PR

**Location**: site-wide

**Why have this**: To catch any broken/outdated links.

**More info**: We can use the GH action like so: https://github.com/OpenPodcastAPI/api-specs/pull/38

PR #256, opened 2023-07-21
Check for valid links on Pull

shinwookim:
Addresses #247 by using [`lycheeverse/lychee-action@v1.8.0`](https://github.com/lycheeverse/lychee-action) in GitHub action `autoreview`.

## Notes
- Can customize what links to ignore for link checking in `./.lycheeignore`
  - Currently set to ignore the AntennaPod forum links as they can time out due to too many requests
  - Currently ignores llama links (in Tasker.md)


.github/workflows/autoreview.yml
```diff
@@ -10,6 +10,15 @@ jobs:
...
         pattern: "*.md"
+  checklinks:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.8.0
+        with:
+          args: --verbose './**/*.md' 
+          fail: false
```
keunes, line 21: Any specific reason you don't fail the action if the link check fails? Would it still give an in-your-face warning? (Just to avoid a PR gets merged even though there are broken links.)

keunes (commented):
Hi @shinwookim,
Thanks for your PR! I was just wondering about the effect of not failing the workflow in case of link issues.

.github/workflows/autoreview.yml
```diff
@@ -1,5 +1,5 @@
 name: autoreview
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
```
keunes, line 2:
```suggestion
on: [pull_request]
```

.github/workflows/autoreview.yml
shinwookim: Just in case the checklinks job completes before the other ones, I didn't want the other jobs (in autoreview) getting cancelled. I figured rather than failing the entire Job, a human reviewer could perhaps look at them instead.

There should still be a summary in actions page that looks like below. But I suppose this can be furthered developed into perhaps creating auto issues in the repo.

image

ByteHamster: My experience is that a check that does not fail is basically useless because one doesn't always open the job output page before merging if there is already a green checkmark

keunes: Ok. @shinwookim, would you mind changing it after all? If I'm not mistaken, only that linkcheck will then fail; it wouldn't prevent other tests in autoreview from running. And if needed, we can still merge PRs despite a failing test.

keunes:
Ok, I see that there are 29 errors ^^ Will have a look at them and address as much as possible via this PR, such that the next person submitting won't have this issue :-)

keunes:
Hi @shinwookim,

Sorry for the delay, and thanks for your contribution!

PR #260, opened 2023-07-30
Refactor footer

Rotzbua:
* Add icons
* Remove useless html tags
* Remove redundant CSS classes
* Fix image accessibility


keunes:
Hi @Rotzbua
Thanks for this. Could I just ask you to revert/remove the adding of the icons? It makes the footer busier. It's common practice to include icons for the 'social channels' but not so much for the website areas (certainly because we don't use them anywhere else in our navigation). (Even though I must admit I'm charmed by the icons for the license and privacy policy pages.)

Thanks!

Rotzbua:
Thanks for your commit. I overlooked that your comment was a change request, which still needs an implementation.

PR #261, opened 2023-07-30
Refactor highlighting current menu

Rotzbua:
* Add CSS
* Remove not working JS code

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -12,8 +12,14 @@
         <div class="collapse navbar-collapse flex-grow-0 weight-500">
           <div class="navbar-nav align-items-center">
             {% for link in site.data.navbar %}
-            <div class="nav-item ml-2">
-              <a class="nav-link" href="{% unless link.url contains 'http' %}{{ site.baseurl }}{% endunless %}{{ link.url }}">{% t {{ link.key }} %}</a>
+            {% assign link_collection = link.url | replace: '/', '' %}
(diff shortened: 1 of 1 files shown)


keunes:
Hi @Rotzbua,

Thanks for this. It seems, unfortunately, that this doesn't work.

Before (live site):
afbeelding

After (Netlify & local build):
afbeelding

---

I think there's three issues:
* With the help of ChatGPT I figured out what the JavaScript is doing, and it adds the 'active' class to `<div class="nav-item ml-2">` rather than the anchor element. So we need to move the if-statement.
* `{{ page.url }}`returns a trailing slash: `/events/`, while `{{ link.url }}` doesn't have one. Guess that can be fixed also.
* The JavaScript (again, I didn't come up with this myself) "_split[s] the path into an array using '/' as the separator. The `[1]` index retrieves the second element of the array, which corresponds to the first part of the path after the initial '/' (typically representing a section of the website)._" In other words: whatever your page URL, it will always come up with the section name, and highlight the parent section. As a specific sub-page under the section (e.g. Events > Community meeting) will have a different page.url (e.g. /events/community-meeting), it won't highlight the parent section when on a child URL. Maybe we can use jekyll `split` then [`slice`](https://shopify.dev/docs/api/liquid/filters/slice) to get the first part of the `page.url`, just like in JS?

Thanks again!

keunes:
Thanks again @Rotzbua!

PR #266, opened 2023-08-08
Update font `Sarabun`

Rotzbua:
Update from v8 to v13.

## Reference

Source: https://gwfh.mranftl.com/fonts/sarabun?subsets=latin


keunes:
I didn't realise fonts had versions ^^ How did/can you see which version we have? (I installed [Font Manager](https://github.com/FontManager/font-manager) but it doesn't seem to display a font version, not for Sarabun at least.)

Rotzbua:
In this case the version is not the font-version. It is the version number from google fonts. This may include some changes or fixes. Sometimes a new version is just a higher compressed release.
I could not find a release-log for google fonts.

keunes:
Thanks for the explanation @Rotzbua. I understand now that it's about the Google font. But then where is the font version declared; how did you know the old one was v8? (Sorry for these annoying questions, I'm trying to understand our website better :P )

Rotzbua:
> how did you know the old one was v8

Filename 😉 : sarabun-**v8**-latin-700.woff2

>  But then where is the font version declared

The version is declared by google. The version is published by their API:

https://developers.google.com/fonts/docs/developer_api

The helper page [gwfh.mranftl.com](https://gwfh.mranftl.com/fonts/sarabun?subsets=latin) just access this API and creates a zip bundle for easier selfhosting.

Google has a public repository where all the magic happens:
https://github.com/google/fonts

The latest changes for `sarabun` can be found here:
https://github.com/google/fonts/tree/f4352e3b2105d6cdf68a06fbd3592fc844454aab/ofl/sarabun

The latest changelog is (5 years ago):

> sarabun: unhinted fonts.
> In the future, we plan to release this family as an unhinted variable font.

tl;dr: it's a rabbithole!
In general: fonts are real science. You can find whole books how to designing fonts 📚 . With modern displays and the pixel and subpixel pattern there are other aspects which are important for fonts....

keunes:
> Filename 😉 : sarabun-v8-latin-700.woff2

Oh, doh 😄

Thanks for the explanations! (And the contributions of course)

PR #268, opened 2023-08-25
Try harder to ask F-Droid people to be patient

ByteHamster:


keunes:
Not sure this will help; it's not very specific. Also it's quite down the page (so unlikely to be read).

I'll have a look at it tomorrow.

PR #280, opened 2023-10-15
Update html heading structure

femmdi:
Structure headings to have only one H1 per page.

### DIFF
--- a/_i18n/en/contribute/develop.md
+++ b/_i18n/en/contribute/develop.md
@@ -1,31 +1,31 @@
 As a community project, AntennaPod relies on voluntary code contributions from a wide range of people. Not just for the app itself, but also the website and related projects.
 
-# 1. App (Java)
+## 1. App (Java)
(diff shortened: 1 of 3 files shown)


ByteHamster:
I wonder if we can shift the headings one level down programmatically. Feels a bit strange to have a markdown file where `#` is not allowed because it is already used outside the markdown file

keunes:
Yeah, was thinking the same. Haven't looked into a technical solution yet though.

ByteHamster:
Apparently it's okay to have multiple `<h1>` on the same page if one of them is in an `<article>` or similar. Then we just need to adapt the font size instead of actually shifting the levels. Given that we are discussing some website changes lately, should we maybe postpone this until the website changes are done?

keunes:
> Given that we are discussing some website changes lately, should we maybe postpone this until the website changes are done?

Yes, I think that's best.

PR #288, opened 2023-11-23
Add collection and ads shortlink

keunes:


keunes:
Hi @Rotzbua,

Could I call on your help here? I want to turn the shortlinks into a collection, so we can
* set default frontmatter for the entities, and
* create a shortlinks 'index' page (so I can see which shortlinks we have, when replying to Play store reviews)

Envisioned approach: https://github.com/AntennaPod/antennapod.github.io/pull/287/files

I created the collection, and I'm using the redirect layout file, which is already used for the 'manually' created pages. The issue seems to be in `_includes/head.html`, because removing this include from `_layouts/redirect.html` solves the issue - it builds successfully. For some reason, there is a null string somewhere. As you can see from the commit history, I tried to find the source of the problem - unfortunately to no avail.

Annoyingly, the site builds with no problem at all locally on my machine (I have different versions than what GitHub uses). So it's hard to get relevant debug information.

Do you have issues building the current state locally? Any idea why we're ending up with a null string somewhere?

PR #295, opened 2023-12-19
fix(accessibility): fix missing alt for img

Rotzbua:
Error: An img element must have an alt attribute, except under certain conditions.
Source: Nu Html Checker


keunes (commented):
Seems I never submitted my review 🤦Sorry!
I would just rename this, as it's the whole website's menu.

keunes:
Alt approach is correct: https://www.w3.org/WAI/tutorials/forms/labels/#labeling-buttons

If we want to make the menu properly accessible, we should probably set the [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) attribute when the menu is open.

PR #296, opened 2023-12-19
fix(html): duplicate id

Rotzbua:
Error: Duplicate ID navbarDropdown.
Source: Nu Html Checker

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -24,12 +24,12 @@
             {% endfor %}
             <div class="nav-item ml-2 ml-sm-1 text-center">
               <div class="nav-item dropdown mb-0">
-                <button class="nav-link dropdown-toggle" id="navbarDropdown" aria-label="{% t generic.accessibility.lang-menu %}"
+                <button class="nav-link dropdown-toggle" id="navbarDropdown1" aria-label="{% t generic.accessibility.lang-menu %}"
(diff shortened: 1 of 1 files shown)


keunes:
What I'm wondering here, is why we have the language dropdown twice in the first place. I believe one is for mobile view, the other 'desktop'. Wouldn't it be better to 'merge' them (code-wise), if possible?

ByteHamster:
Is this maybe also something we could change with the new website design?

Rotzbua:
@keunes Yes there are two language dropdowns. One for mobile and one for desktop. Since it is a very simple fix, I see no reason to wait until a future solution (refactor, new design) implement it.

keunes:
> Since it is a very simple fix, I see no reason to wait until a future solution (refactor, new design) implement it.

Fair enough.

@ByteHamster I agreed initially that this could be addressed with the website update. But we're probably going to get raw HTML+CSS and would have to translate it into Jekyll+SASS ourselves (if I understood correctly). Meaning we might just take/implement the CSS changes and update the Jekyll includes/templates (rather than reconverting from scratch). In that scenario we'd probably have to refactor the mobile menu separately.

@Rotzbua Would it be feasible for you to refactor the language menu now? If yes, I think that'd be great. If not, can you please force push to resolve the conflict?

PR #297, opened 2023-12-19
chore(jquery): update to 3.7.1

Rotzbua:
https://code.jquery.com/jquery-3.7.1.min.js


keunes:
Thanks. I'm on my work laptop and don't remember from the top of my head: do you know where/what for we use jquery? Just to make sure it's all working as expected still.
(If not, I'll have a look some point next week, after Christmas.)

ByteHamster:
Why do we need to upgrade it anyway? In my experience, unless there is an actual reason to upgrade dependencies, upgrading them just makes a headache

keunes:
> dependencies, upgrading them just makes a headache

I can understand for mobile apps where there's a lot of different contexts, where you can't test them all yourself.

But for my laptop, having to jump major software versions & their dependencies actually gives a headache. Oftentimes I find myself starting from scratch because it's easier.

Same for web development I think: we can actually test all major cases ourselves; the list of variables is much smaller.

Rotzbua:
jQuery is used by Bootstrap 4.

PR #330, opened 2024-02-25
Ignore link on link checker

ByteHamster:
Uses cloudflare redirects

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -5,3 +5,4 @@ https:\/\/twitter.com/intent/tweet.*
 https://www.pexels.com/@.*
 https://help.soundcloud.com/.*
 https://pixabay.com/de/users/.*
+https://www.ethnologue.com/guides/ethnologue200


keunes:
@ByteHamster I think these links should be updated still? The links are effectively changed (even though they redirect). I think I updated the source, but it might not have been updated in all languages yet.
Current URL: https://www.ethnologue.com/insights/ethnologue200/

ByteHamster:
Ooh, that's what is going on. I somehow missed that. Then I guess we should revert this PR?

keunes:
I suspect so, yes.

Also, about this one:
Errors in _site/es/blog/2022/10/2-7-release.html
    [403] https://pixabay.com/de/users/anne-onyme-1513318/

Fine to exclude pixabay & pexels, but this (and similar) should be corrected still (es AntennaPod site links to de pixabay site)

keunes:
All localised links updated on Weblate. If you revert this and we pull updated translations before the next link checker run, ethnologue should no longer cause trouble.

ByteHamster:
Reverted and pulled the translations :) Let's see if the link checker runs through

PR #332, opened 2024-03-03
Update Subscription page javascript

### LINKED ISSUE #331: 99% invisible share page not rendering correctly
**Short description**: When tapping 'share' in AntennaPod from the '99% invisible' podcast, the 'Subscribe' page (based on the created URL) doesn't render correctly: It doesn't include the Podcast's title & the field that should display the RSS feed says 'JavaScript required'.

**URL**: `https://antennapod.org/deeplink/subscribe/?url=https%3A%2F%2Ffeeds.simplecast.com%2FBqbsxVfO&title=99%25+Invisible`

**More info**: Introduced in #303 & https://github.com/AntennaPod/AntennaPod/pull/6835. It seems that the problem is the fact that 99 & the percentage symbol in the title are glued together. If a space is added between them the page renders correctly.

CC: @ByteHamster

keunes:
Fix podcast data decoding issue & display friendlier message if JavaScript not available.

Closes #331 (disclaimer: solution proposed by ChatGPT)

### DIFF
--- a/deeplink/subscribe/index.html
+++ b/deeplink/subscribe/index.html
@@ -14,7 +14,7 @@ title: generic.subscribe
             <div class="card">
               <div class="card-body">
                 <i class="fa fa-rss mr-1"></i>
-                <span id="urlTextBox">JavaScript required</span>
+                <span id="urlTextBox">To display dynamic data about the shared podcast, please ensure JavaScript is active</span>
               </div>
             </div>
           </p>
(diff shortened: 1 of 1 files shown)


ByteHamster:
This solution is a bit hacky ;) I already wrote a more elegant solution but you were a bit faster pushing (currently struggling with the train wifi) :) See  #333

PR #353, opened 2024-07-13
Remove gpodder from privacy policy

ByteHamster:
We no longer promote the gpodder server and no longer support searching.

The link often fails anyway, see #348


_i18n/en/general/privacy.md
```diff
@@ -15,8 +15,8 @@ The app requests the following privacy-relevant Android permissions:
...
-- Synchronisation services: When enabled via the settings, AntennaPod synchronises your data. This data may include login credentials, subscribed podcasts, listened episodes, play, pause & favorite actions with date & time stamps and IP address. For more information, see your provider's privacy policy (e.g. [gpodder.net's privacy policy](https://gpodder.net/privacy), or the one of your Nextcloud host).
+- Discovery and search services: When opening the Discover screen in the app, it sends a call to Apple to collect podcast suggestions. They may store the query, including country (which by default is the device's country). These suggestions can be disabled by choosing "off" in the region selector. When using the search feature in AntennaPod, the services PodcastIndex.org, Apple Podcasts, and fyyd may store your query, including the search terms. Discovery and search queries also include IP address, time, and app name "AntennaPod". After subscribing, the feed is served from the podcast hoster, meaning the discovery/search service is no longer involved.
```
keunes, line 18:
```suggestion
- Discovery and search services: When opening the Discover screen in the app, it sends a call to Apple to collect podcast suggestions. They may store the query, including country (which by default is the device's country). These suggestions can be disabled by choosing "off" in the region selector. When using the search feature in AntennaPod, the services PodcastIndex.org, Apple Podcasts and fyyd may store your query, including the search terms. Discovery and search queries also include IP address, time, and app name "AntennaPod". After subscribing, the feed is served from the podcast hoster, meaning the discovery/search service is no longer involved.
```

PR #356, opened 2024-09-23
Hacktoberfest 2024 blog

keunes:
Hoping that absence of image and author will be OK.


ByteHamster:
Hmm, missing image and author doesn't look too nice. I would just use some random stock photo

keunes:
> missing image and author doesn't look too nice. I would just use some random stock photo

Agree. But I'm on mobile so it's a bit too complicated to address all. Hence the request on the forum for someone else to fix it 😇

PR #357, opened 2024-09-25
Make lead paragraph look more consistent

ByteHamster:
Before: So many different font styles on one single screen
grafik

After: Number of font styles reduced by 1
grafik

### DIFF
--- a/_sass/_custom.scss
+++ b/_sass/_custom.scss
@@ -1220,3 +1220,10 @@ footer .svg-wave {
     display: none !important;
   }
 }
+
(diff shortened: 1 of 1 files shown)


loucasal:
I'm afraid the hardcoded background colour doesn't adapt well to dark mode - see screenshot.

Screenshot 2024-09-29 at 23-38-06 AntennaPod is participating in Hacktoberfest 2024 – Join us! – AntennaPod

ByteHamster:
Thanks, fixed in #361

PR #364, opened 2024-10-11
other-platforms.md add about Genymotion, Waydroid, Bliss OS

vitaly-zdanevich:


ByteHamster (commented):
Thanks, I added some comments below

vitaly-zdanevich (commented):
Solve the feedback - move to introduction, drop BlissOS

vitaly-zdanevich (commented):
Typo of -> or

vitaly-zdanevich:
Ok, but if somebody tries to port to Linux - is it difficult? Where to start?

keunes:
Hi @vitaly-zdanevich
Thanks for contributing to AntennaPod!

To be honest I'm not convinced this should be merged. It's complicated for the average user to do (while those really interested in it will be able to find these tools via a web search or ask us on the forum), it's currently unclear what these do/what these are, at least Genymotion is not open source, and I'm a bit afraid that if we start recommending installing & running AntennaPod in unofficial ways we might get related support questions.

WDYT @ByteHamster?

ByteHamster:
Hmm, right. I do have to agree that running an Android emulator just to use AntennaPod on the desktop probably is not a nice user experience. So we maybe shouldn't recommend this on the website.

keunes:
Thanks @vitaly-zdanevich for thinking of us and preparing a pull request! As discussed, this probably isn't fit for our main website audience, so closing.

PR #367, opened 2024-10-15
Add smart shuffle documentation

### LINKED ISSUE #180: Document working of Smart Shuffle
**Short description**: Add a documentation page about how smart shuffle works.

**Location**: https://antennapod.org/documentation/queue

**Why have this**: The term 'Smart shuffle' sounds cool but it's not clear at all what it does/how it works.

**More info**: https://github.com/AntennaPod/AntennaPod/pull/6030

loucasal:
Closes #180.

I briefly considered going into the technicalities of the algorithm, but that seems entirely unnecessary for the average user.


ByteHamster:
Thanks!

> I briefly considered going into the technicalities of the algorithm, but that seems entirely unnecessary for the average user.

Then I maybe wouldn't mention Round-Robin. Round-Robin is technically the explanation of an algorithm that we do not use. Users who know what Round-Robin is would then assume that something is broken.

loucasal:
Done - even simpler now.

ByteHamster (approved):

loucasal:
@keunes any comments or can this be merged as it is?

keunes:
@loucasal I'm wondering if it's worth adding a brief section how how it works in detail?

For example:
> Let's say you have 15 episodes in your queue; 1 from podcasts `A`, `B` and `C` each, 2 from podcast `D` and 10 from podcast `E`. It first  It first spreads the podcast with most episodes (`E`) and places them spread out in the queue: `EE_EE_E_E_EE_EE`. The podcast with the second-most number of episodes (`D`) is then placed spread-out in the *available* slots: `EE_EEDE_EDEE_EE`. This continues, until we end up with: `EE B EE D E C E D EE A EE`.
> The publishing date of each episode plays a role a) within podcasts (the episodes of podcast `E` are sorted accordingly) and b) between podcasts (in our example, the episode of podcast B gets an earlier spot than the episode of podcast A, because the former is more recent). Within date sorting, recently published gets higher in the queue.

If not, I think I would add a link to AntennaPod/AntennaPod/pull/6030 for a detailed explanation.

loucasal:
As I said above:
> I briefly considered going into the technicalities of the algorithm, but that seems entirely unnecessary for the average user.

Plus it would be another long chunk of rather detailed text for translators to localise. So let's add a link to the PR instead.

ByteHamster:
I think I would slightly prefer stating the explanation on the website. The reason is that the PR explains the old variant first, which might confuse users. Also, leading them away from the documentation page might be a bit unexpected. Maybe we could add it with a new heading and have it excluded from the translations?

loucasal:
Ok, I will address this over the weekend so we can finally merge this.

keunes:
> Maybe we could add it with a new heading and have it excluded from the translations?

New heading; yes, that's what I proposed. Excluding it from translations; that's odd. I agree it's not a priority, but having just one paragraph on one page not translated looks a bit weird.

If we want to save translators some work, we should consider disabling the whole 'contribute' component (because that's lowest priority since we expect contributors to understand English, while translators often do that still before documentation).

loucasal:
@keunes @ByteHamster I have reworked the entire description in the latest commit.

Two comments:
- I omitted the last sentence because it was unclear to me and possibly redundant. Does 'within date sorting' mean 'when sorting episodes that have the same publication date'? If so, I think the concept is already covered in the previous sentence.

- I must say that when fact-checking this description, I was extremely confused by the fact that the sort order is _descending_ by default. This affects both the "smart shuffle" and "date" sorting criteria. Wouldn't the base assumption be that users are more likely to listen to episodes in chronological order (rather than reverse chronological order)? I can't see that this was discussed the last time these settings were discussed/reworked in antennapod/antennapod#6785.

keunes:
> Does 'within date sorting' mean 'when sorting episodes that have the same publication date'?

Maybe 'within' is the wrong word. What I meant, was 'in the context of date sorting' we're doing 'newest to oldest' (not oldest to newest). But I guess indeed it's fine to leave that sentence out.

> I must say that when fact-checking this description, I was extremely confused by the fact that the sort order is descending by default. This affects both the "smart shuffle" and "date" sorting criteria.

I hadn't tried out the smart shuffle properly actually, when writing the text. I agree with your comment, although the pure 'date' sorting does the same thing: triangle down (the default) means from new to old. One could argue about whether 'triangle down' represents 'new to old'. But I agree that it's not what I would expect the default.

loucasal:
Filed antennapod/antennapod#7519 regarding the change in the default sort order. Depending on how quickly that is taken up, we can either merge this as is and revisit it later, or keep it open until that issue is addressed.

ByteHamster (approved):
I would merge as it is. We don't know how soon we might change the default order.

keunes:
Thanks both!

Issue #311, opened 2024-01-02
Contributor documentation


keunes:
**Short description**: Add contributor documentation.

**Location**: 

**Why have this**: 

From https://github.com/AntennaPod/AntennaPod/pull/6834#discussion_r1439050615:
> It would be nice if there was an easy way to auto-format it according to the requirements but I didn't see anything in the contributor guide

The contributor's page for 'Develop' links to this wiki page: https://github.com/AntennaPod/AntennaPod/wiki/Code-style. Apparently it's not visible enough.

Also, the Contribute > Translations page is rather long and complicated.

**More info**: 

Pages to include:
* Requesting new languages
* Using Weblate
* Code style
* Building AntennaPod
* About debug versions

keunes:
> The comment linked above doesn't imply that they did not find the developer documentation. It just says that the developer documentation does not contain what they were looking for (now it does).

I didn't realise you edited the wiki page to add the info :-)

> I'm still not really convinced that the public website should hold very technical stuff like the developer documentation

It doesn't have to be in the menu, confusing 'regular' users. But as a) important information for contributors is currently spread over different places and b) hard to process (at least the translations), we need to do _something_ to address both issues.

And as our main website goal is to attract contributors and ensures a good mix of controllability and openness (using PRs, impossible combination on the wiki), I find it the most appropriate place. (But if you have another proposal that addresses identified issues, I'm all ears.)

ByteHamster:
Was your idea to completely remove the wiki and the `CONTRIBUTING.md` file from the main app repo to have everything in one place?

keunes:
My idea was indeed to drop the wiki entirely. CONTRIBUTE.md probably not, as it's a recognised standard. However, if we have a contributor's section, we should limit it to key information & pointers.

ByteHamster:
If we do this, I would limit the `CONTRIBUTING.md` file to just contain a link to the website. Then we really have everything in one place. Otherwise we move from wiki+repo to website+repo, so the information would still be spread over different places.

I assume the contributor documentation won't get translated, right? Translating variable naming guidelines and stuff doesn't make a lot of sense because the variables themselves should be in English anyway. Also, contributions are discussed in English as well.

keunes:
>  I assume the contributor documentation won't get translated, right?

Yeap, agree.

loucasal:
I could help with (at least parts of) these documentation pages next, starting with the translation bits.

However, I have a few questions before I set myself to the task:
[...]

PR #372, opened 2024-11-30
Split develop page into multiple pages

loucasal:
Prerequisite to adding further content as part of #311.

This is less straightforward than adding pages to existing sections, so feedback appreciated.


loucasal:
I'm a bit stuck with this one. I don't understand why the `intro` string in `_i18n/en.yml` works for Documentation pages, but not for Contribute ones. @keunes, can you see where the problem may be?

keunes:
Hi @loucasal,

Sorry for the late reply. It's a bit busy here. One issue is that `contribute.` is missing when you're calling the strings. For example in `/contribute/develop/index.html`:

```md
---
layout: internal
title: contribute.develop.title
excerpt: contribute.develop.text
```
Here you weren't referencing strings that existed in `/_i18n/en.yml`, specifically the top-level 'contribute' indication is missing.

Also, the files of the individual sub-pages are missing under `/contribute/develop` (e.g. `/contribute/develop/app.md`), similar to `/contribute/develop.md`. It is a bit confusing, but for each page you need an md or html file (to generate the page) _and_ an md file containing the actual text (to provide the text for translations).

If you address these two points, that should get you further I think.

loucasal:
I can't get this to build even after trying to address your suggestions. I'm tempted to leave this to someone who can deal with this more efficiently...

keunes:
I'll see if I can get it to work @loucasal.

loucasal:
Thanks @keunes! I wanted to get this done to be able to build on it for other parts of #311, so you can merge away if it works.

loucasal:
@keunes anything holding this up? Let me know if I can help.

keunes:
@loucasal Time was holding this up :-) I started a new PR (#390) because that was a bit easier/cleaner for me. I managed to turn the Contribute pages into a collection, from there we can make further changes. It's currently not building due to the translation files still including some strings that we have removed. I'll address that as well by updating the translation files, so it will build tests will pass (after which we can make further changes).

PR #377, opened 2024-12-15
Explain Google Chromecast

ByteHamster:


keunes:
I'd use the Chromecast icon, if we can: https://fontawesome.com/icons/chromecast?f=brands&s=solid

PR #379, opened 2024-12-26
Clarify again that there is no need to ping us about F-Droid releases

ByteHamster:
- Explicitly state that "Play Store shows date X" does not mean that it should already be on F-Droid (staged rollout).
- Even more explicitly tell users that there is really no need to ask about the F-Droid releases every single time.
- Make first paragraph a lot shorter so they really read the main message: "be patient!"


keunes:
I'm thinking of some further changes. Will push later this week.

keunes:
I'm thinking of some further changes. Will push later this week.

PR #385, opened 2025-01-04
Fix empty strings to translate

ByteHamster:
Introduced in a9cb33e87d8ffa2a5ad65176ee6ade1ad2b37e95


keunes:
@ByteHamster These empty strings were introduced to allow the site to be built (same issue as #391). The texts that you added here are wrong, though: the three screenshots on the homepage are exactly the same and don't show a specific functionality, as the texts now suggest. Therefore I would be inclined to revert (or, if #391 enables us, remove these strings altogether).

PR #390, opened 2025-02-09
Update Contribute section

keunes:
Closes #372, prepares the ground for #311

- [x] Turn Contribute pages into collection
- [x] Split develop pages
- [x] Make 'donate' a regular card
- [ ] Move translations to the new structure
  - It will allow to publish the existing translations faster, but given our system with a `translation-files` branch, I don't think it would push those translations back to Weblate and avoid copy-paste work by translators. To be checked if updating the translation files branch would help. During experimentation we should lock the resources on Weblate.
- [x] Merge 'contribute' and 'documentation' cards
- [x] ~~Split Translate into sub-pages~~ Revise and shorten Translate page


ByteHamster:
Hmm, I wouldn't link directly to OpenCollective. We really don't need the money. If we make it easy to donate, people just donate money and leave. We should really try to get them to contribute instead. (note also that it is not technically a donation, maybe we should phrase it differently?)

If we really have to make it more prominent than the small link, I would open a page that says something like the popup we already have. That popup also got a lot of nice feedback on social media ;)

ByteHamster:
> Merge 'contribute' and 'documentation' cards

What do you mean with that? Unifying the design?

keunes:
> Hmm, I wouldn't link directly to OpenCollective.

It still opens the modal with the 'warning'. As Ryan mentioned, if a user is looking for a 'donate' button because that's the only way they can contribute, hiding it will not do any good.

To funder improve the situation I hope that we can get https://github.com/AntennaPod/AntennaPod/issues/5153 sooner rather than later, and add a button on the website that will open the relevant screen directly.

>> Merge 'contribute' and 'documentation' cards
>
> What do you mean with that? Unifying the design?

The design is already pretty much unified. The only difference is the presence of an explanation text. I want to deduplicate the code (without changing the design further).

keunes:
@loucasal If you want to have a look and share any feedback - I now have a workable version that can be checked here: https://deploy-preview-390--antennapod-test.netlify.app/

_i18n/en/contribute/translate.md
```diff
@@ -1,28 +1,55 @@
...
+Out of the estimated [world population](https://en.wikipedia.org/wiki/World_population) of 8 billion people, 'only' 1.5 billion [speak English](https://www.ethnologue.com/insights/ethnologue200/). This may not mean much per se. But if anything, we can conclude that making AntennaPod available in more languages, makes the app accessible to more people. As a side benefit, localizing the app also helps our project grow beyond the current main userbases (Germany and the US).
 
 We therefore much appreciate your help!
 
-To help translate the app, we ask you to contribute via [Weblate](https://hosted.weblate.org/projects/antennapod/app/). Anyone can sign up there and get started. In Weblate's documentation you'll find [everything you need to know to start translating](https://docs.weblate.org/en/latest/user/translating.html).
+Our translations are hosted on Weblate, an open source translations platform. Anyone can sign up there and [start translating for AntennaPod](https://hosted.weblate.org/projects/antennapod). You'll find everything you need to know in [Weblate's documentation for translators](https://docs.weblate.org/en/latest/user/translating.html){:target="_blank"}.
```
keunes, line 5:
```suggestion
Our translations are hosted on Weblate, an open source localisation platform. Anyone can sign up there and [start translating for AntennaPod](https://hosted.weblate.org/projects/antennapod). You'll find everything you need to know in [Weblate's documentation for translators](https://docs.weblate.org/en/latest/user/translating.html){:target="_blank"}.
```

_i18n/en/contribute/translate.md
```diff
@@ -1,28 +1,55 @@
...
 
-New translations are pulled from Weblate to GitHub (the development platform) at least for every 'point release'. These are releases like 2.0, 2.5, etc that introduce new features. But we also release updates with bug fixes and new translations in between. On average, it might take a month or two before your work is available in a release.
+<object data="https://hosted.weblate.org/widget/antennapod/horizontal-auto.svg" type="image/svg+xml" width="100%" height="auto" crossorigin="anonymous">
+<a href="https://hosted.weblate.org/engage/antennapod" target="_blank">{{- img-translationstatus | strip -}}</a>
+</object>
+<!-- mdpo-enable -->
 
-Translation contributors will find themselves mentioned in the app with their Weblate username, under `Settings` » `About` » `Contributors`.
+## 1. The app
+The app is currently available in more than 38 languages. While there are more languages in Weblate, we only include those languages in the app which are at least 40% translated. Below that threshold the app hardly provides a good user experience.
```
keunes, line 28:
```suggestion
The app is currently available in more than 38 languages. While there are even more translations in Weblate, we only include languages which are at least 40% translated. Below that threshold the app hardly provides a good user experience.
```

_i18n/en/contribute/translate.md
```diff
@@ -1,28 +1,55 @@
...
+## 2. The website
+The website exists of four 'components', each representing a part of the website. We encourage you to follow this order:
 
-# Get in touch!
+1. **General** and **Short strings**: the first elements a user sees on the website
+2. **Documentation**: our support materials, which are directly available via the app
+3. **Contribute** (comes last because a basic understanding of English is needed anyway in order to contribute)
+
+Updated translations are published once in a while. New languages are enabled on the website only after priority 1 & 2 components are fully translated. Please let us know through the forum (see below) if those components are done. We'll then try to find a community member who can review a test version of the website, to improve the quality of translations. After the review, we can publish the new language version.
```
keunes, line 41:
```suggestion
Updated translations are published once in a while. New languages are enabled on the website only after priority 1 & 2 components are fully translated. In other words: when everything but the contribution section is done. Please let us know through the forum (see below) if that's the case. We'll then try to find a community member who can review a test version of the website, to improve the quality of translations. After feedback is processed, we can publish the new language version.
```

_i18n/en/contribute/translate.md
```diff
@@ -1,28 +1,55 @@
...
+
+## 3. The app store listings
+In addition to the app and website, we also translate texts which appear in the various app stores. We encourage you to follow this order:
+
+1. **Short description**
+2. **Full description**
+3. **Screenshot titles**
+
+We have to comply with a character limit for the short descriptions and screenshot titles. We therefore manually review improve these texts before they are pulled and made available. Please let us know on the forum once you've translated these texts.
```
keunes, line 50:
```suggestion
We have to comply with a character limit for the short descriptions and screenshot titles. We therefore manually review and improve these texts before we publish them. Please let us know on the forum when you have completed translations of these texts.
```

keunes (commented):
Some further tweaks to the Translate texts.

loucasal:
Looks fine to me, but I didn't review the code. I just have a few minor suggestions on texts.

_i18n/en/contribute/translate.md
loucasal:
```suggestion
The app is currently available in more than 38 languages. While there are even more translations in Weblate, we only include in the app languages which are at least 40% translated. Below that threshold the app hardly provides a good user experience.
```

It seems clearer to still mention "in the app".

_i18n/en/contribute/translate.md
loucasal:
```suggestion
Updated translations are published once in a while. New languages are enabled on the website only after priority 1 & 2 components are fully translated. In other words: when everything but the `Contribute` section is done. Please let us know through the forum (see below) if that's the case. We'll then try to find a community member who can review a test version of the website, to improve the quality of translations. After feedback is processed, we can publish the new language version.
```

keunes:
> I just have a few minor suggestions on texts.

Did you share them somewhere and I missed them, or did you start a review but not submit it yet? 🙂

loucasal:
They should be visible directly below some of your suggested texts.

keunes:
Ah I see. Didn't get a notification about the replies. Thanks, will have a look tomorrow afternoon.

_i18n/en.yml
```diff
@@ -102,98 +102,143 @@ readtime:
...
-      playback-stops: "Playback suddenly stops"
-      refresh-download-errors: "Download or feed refresh error"
-      cannot-resume-playback: "Can't resume playback from lock screen / notification"
-    podcasters-hosters:
-      title: "For Podcasters & Hosters"
-      intro: "Below, you can find our support information for podcasters & hosting providers."
-      add-on-antennapod: "Create an 'Open in AntennaPod' link"
-      list-podcast: "Getting your podcast listed"
-      update-feed-urls: "Updating feed URLs"
+  getting-started:
```
ByteHamster, line 148: What happens with the translations after this restructuring? Do they all need to be re-done?

_includes/doc-card.html
```diff
@@ -0,0 +1,24 @@
...
+            {{- include.item.external-link }}
+        {%- else %}
+            {%- if site.lang != 'en' %}/{{ site.lang }}{% endif %}
+            {%- if include.item.internal-link %}
+                {{- include.item.internal-link }}
+            {%- else %}
+                {{- include.item.url | replace:'.html','' }}
+            {%- endif %}
+        {%- endif -%}
+        "{% if include.item.slug == "donate" %} data-toggle="modal" data-target="#donationModal"{% endif %}>
```
ByteHamster, line 13: How about moving the donation warning message to its own page? Then we don't need the special handling here. Things like this make it hard to maintain code because the caller has a hard time seeing what happens inside.

_layouts/doc.html
```diff
@@ -1,6 +1,11 @@
 <!DOCTYPE html>
 <html lang="{{ site.lang }}">
 <head>
+{%- if page.external-link -%}
+  <meta http-equiv="refresh" content="0; url={{ page.external-link }}">
+</head>
+</html>
+{%- else -%}
```
ByteHamster, line 8: Couldn't this be achieved a bit more locally by setting the layout of the docs sub-page to "redirect"? Having this in the global "doc" layout is a bit unexpected from a software architecture point of view. Things like this make it harder to maintain in the future.

Gemfile.lock
```diff
@@ -54,7 +54,7 @@ GEM
     jekyll-mentions (1.6.0)
       html-pipeline (~> 2.3)
       jekyll (>= 3.7, < 5.0)
-    jekyll-multiple-languages-plugin (1.8.0)
+    jekyll-multiple-languages-plugin (1.7.0)
```
ByteHamster, line 57: The lockfile is generated automatically and not really intended to be edited manually. I would rather configure this from `Gemfile` like this:

```
gem "jekyll-multiple-languages-plugin", "1.7.0"
```

Then it ensures that the version stays the same even when updating components with `bundler update`

_i18n/en.yml
keunes: To be confirmed

_includes/doc-card.html
keunes: From a UI perspective I prefer to keep it in a modal.

_layouts/doc.html
keunes: Right. I guess we could, but with the current set-up I believe the user would see a flash of the redirect page, which isn't great for internal purposes. So if we take that approach, I would remove the delay (if it's there).

ByteHamster: There is no delay. The redirect page uses the exact same meta tag that you added here now.

_includes/doc-card.html
ByteHamster: On a page, we could show things like a button to open subscriptions with donation links, explain why it is important to support podcasters to keep the ecosystem open, etc

_layouts/doc.html
keunes: Then I don't know why I remember seeing flashes of the Redirect page. Will update this accordingly.

_i18n/en/contribute/translate.md
keunes:
```suggestion
The app is currently available in more than 38 languages. While there are even more translations in Weblate, we only include languages in the app which are at least 40% translated. Below that threshold the app hardly provides a good user experience.
```

_i18n/en/contribute/translate.md
keunes:
```suggestion
Updated translations are published once in a while. New languages are enabled on the website only after priority 1 & 2 components are fully translated. In other words: when everything but the 'Contribute' section is done. Please let us know through the forum (see below) if that's the case. We'll then try to find a community member who can review a test version of the website, to improve the quality of translations. After feedback is processed, we can publish the new language version.
```

contribute/index.html
```diff
@@ -36,4 +31,3 @@ <h5 class="modal-title" id="donationModalLabel">{% t contribute.donate.modal-tit
       </div>
     </div>
   </div>
-</div>
```
ByteHamster, line 39: Are you sure this line should be deleted?

_includes/doc-card.html
ByteHamster: Needs decision meeting comment: 

Add here:
```
<div class="col-12 col-md-3 mb-3 mb-md-3" id="card-{% include.item.slug %}">
```

On contribute page (above the popup layout):

```
$("#card-donate").attr("data-toggle", "modal").attr("data-target", "#donationModal")
```

keunes:
@ByteHamster I think all addressed. Ready to merge?

ByteHamster:
What about the translations restructuring? This would mean we essentially lose half of the website translations. I would rather keep using the old string structure rather than "building" the string keys dynamically if we can keep the translations that way. (actually, I would say that restructuring the string keys would better fit into an independent PR, as it is not about the contribute page)

keunes:
Right. So I just had a look in Weblate, and I propose to enable the Automatic translation add-on for the `website/short-strings` component with the settings:
* Mode: Add as translation (not suggestion)
* Search filter: Untranslated strings
* Source: Machine translation > Weblate Translation Memory
* Score threshold: 100%

That should automatically fill again all strings which are identical as our past translations.

I really prefer not to go back to manual setting of keys in files, as it makes it more complex to manage content, therefore makes the system more error-prone.

ByteHamster:
> That should automatically fill again all strings which are identical as our past translations.

Okay, if you are confident that this works, go ahead :sweat_smile:

PR #391, opened 2025-02-23
Downgrade multiple-languages-plugin

keunes:
Downgrade multiple-languages-plugin (See https://github.com/kurtsson/jekyll-multiple-languages-plugin/issues/201)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -54,7 +54,7 @@ GEM
     jekyll-mentions (1.6.0)
       html-pipeline (~> 2.3)
       jekyll (>= 3.7, < 5.0)
-    jekyll-multiple-languages-plugin (1.8.0)
+    jekyll-multiple-languages-plugin (1.7.0)
       jekyll (>= 2.0, < 5.0)
     jekyll-sass-converter (3.0.0)
       sass-embedded (~> 1.54)


ByteHamster:
Hmm didn't the plugin version work before? My guess would be that it is a different problem. Does downgrading in your other PR help?

keunes:
The issue in my other PR is the following:
```liquid
            {%- assign card-text-key = include.item.texts | append: ".card-text" %}
            {%- capture card-text %}{%- t card-text-key %}{%- endcapture -%}
            {% if card-text != empty %}
            <p class="text-main">{{ card-text }}</p>
            {%- endif %}
```
This is the code in `_includes/card.html`, where we:
1. dynamically build the path of the text description, e.g. `contribute.develop.app.card-text` (I do this because I want to have a single definition in the front matter, and want to build paths like for title, intro & card-text from it, rather having to specify each of these paths individually in front matter)
2. capture the actual, translated card-text string. It may or may not be present in the translation file(s). If it does not exist, it returns null
3. if the translated string exists, then display it

The issue exists in step 2: if the string does not exist and it returns null, then it causes the error rather than simply not displaying anything:
> Liquid Exception: no implicit conversion of nil into String in /home/runner/work/antennapod.github.io/antennapod.github.io/_layouts/doc.html

I don't have this issue locally (I have with jekyll-multiple-languages-plugin 1.7.0), and the referenced issue notes that the issue started with 1.8.0 (which is in the Gemfile.lock). Therefore I assume that indeed this would fix the issue I have with my PR (I will test in a moment).

keunes:
In the long run we should probably switch to https://github.com/untra/polyglot, given that multiple-languages-plugin is no longer maintained.

PR #394, opened 2025-03-04
Remove alt text strings

keunes:


ByteHamster:
That sets the texts to an empty string. Instead, I would remove them completely

keunes:
This works actually; I already had pushed the changes to the translation-files branch and on to Weblate, and the strings are gone from there. The difference with before, is that now there's no `""`.

PR #401, opened 2025-04-24
Add Persian

ByteHamster:


keunes:
Nice!
A bit nitpicking, but it seems that the icon paddings/margins of the social icons in the footer and mobile menu and the globe for the Lang switcher are lost. We should probably change that from 'left' to 'start' margin or sth?

PR #430, opened 2025-09-24
docs: automation: correct information in refreshing-podcasts.md

james-smith-za:
In this commit I correct a small inconsistency between what's in the documentation and what I encounter in the actual app. I'm on 3.9.0f downloaded from F-droid.


keunes:
Thanks @james-smith-za !

PR #433, opened 2025-10-04
Fix credit and wording in readme

loucasal:
We weren't properly crediting a contributor and the whole sentence was a bit messy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This repository contains AntennaPod's project website, which is live on https://
 
 ## About the site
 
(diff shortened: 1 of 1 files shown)


keunes:
Hmm. I believe that usernames should still be linkified automatically by Jekyll? (As per #45)

We maybe should investigate that a bit further.

ByteHamster:
This PR is about the readme file on GitHub, not the actual website

keunes:
Whoops, sorry.

PR #436, opened 2025-10-04
Move infrastructure info from wiki to contribute

loucasal:
Contributes to #311.

Omitted as obsolete: Amazon App Store, GitHub Wiki, Google Groups, Transifex.

### DIFF
--- a/_i18n/en/contribute/develop/infrastructure.md
+++ b/_i18n/en/contribute/develop/infrastructure.md
@@ -1,0 +1,62 @@
+This page describes what services and accounts are in use for AntennaPod. The goal is to make it clear who has which passwords and keys.
+
+# App distribution
+- F-Droid
+  - Automatic updates from GitHub tags
+  - F-Droid's signing keys
+- Google Play
+  - Developer account owned by project Gmail account, recovery email address is project email address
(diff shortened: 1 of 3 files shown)


_i18n/en/contribute/develop/infrastructure.md
```diff
@@ -0,0 +1,62 @@
...
+# Email
+- `info@`
+  - Managed by ByteHamster and keunes
+  - Used for the required contact address on Google Play
+  - Auto responder tells users to write on forum or GitHub instead
+- `@forum.antennapod.org`
+  - Managed by ByteHamster (catch-all mailbox)
+  - Used by the forum, checked every 5 minutes
+    - Allows to post+reply via email
+- Project email address (not public)
```
ByteHamster, line 37: Maybe it could be good to give the reason here

```suggestion
- Project email address (not public to avoid spam)
```

ByteHamster (commented):
- I'm still not super convinced that this is something that should be on the website, but if you two really want to, I'm fine
- I would still mention the fact that we don't have access to the other stores. That makes it clear to possible future maintainers (not that I would plan to leave) that we can't update the app there

PR #439, opened 2025-10-05
Improve open in AP page

### LINKED ISSUE #404: Expand article on creating an 'open in AntennaPod' url
**Short description**:
* [ ] Currently the page explaining how to get the 'open in AntennaPod' button does not show how convenient it will be/how it will look for users who have AntennaPod installed. It'd be great if we can show with two screenshots what it'll look like for users who have and who have not installed AntennaPod.
* [ ] The link to the branding materials is a bit too small - it should probably get its own sub-heading
* [ ] Special characters like ampersand (&) must be url-encoded for it to work. That's currently not explained on the page; we should add that.
* [ ] Nice to have: example code for an 'open in AntennaPod' button + rendered button.

**URL**: https://antennapod.org/documentation/podcasters-hosters/add-on-antennapod

**More info**: https://mastodon.social/@russell@podcastindex.social/114425443938214851

loucasal:
Closes #404.

I will need some help completing this one.

- [x] Instructions on URL encoding special characters are included.
- [x] Three screenshots are added: two illustrating the result for users who have AntennaPod installed and one for those who don't.
- [x] The link to the branding repository is moved to a sub-heading.
- [x] I included example HTML code for an 'open in AntennaPod' link. However, even reproducing the button we use on the deeplink page would require showcasing more code, including a script. If the current example code is too basic, it could be improved (help wanted) or left out.

I also thought it would be nice to take the opportunity for two related improvements (help wanted): 
- [] The RSS feed on the [deeplink page](https://antennapod.org/deeplink/subscribe/?url=https://antennapod.org/rss.xml&title=Blog+Posts) contains a leading space which should be removed.
- [] It would be nice to add a clickable "Copy to clipboard" button to the right of the RSS feed on the deeplink page.


loucasal:
Right, the copy to clipboard button is clearly beyond my reach, but at least I managed to resolve the other issue, so that's enough for me.

Issue #337, opened 2024-03-12
Issues with formatting in source/localised pages


loucasal:
**Short description**: 
I spotted some issues with bullet point/numbered lists on at least two documentation pages.

**URL**: 
1. A page where the source page is displayed correctly, but localised pages have issues with bullet point/numbered lists:
https://antennapod.org/fr/documentation/bugs-first-aid/database-error

2. A page where both the source and the localised pages have issues with numbered lists:
https://antennapod.org/documentation/general/synchronization

keunes:
> A page where both the source and the localised pages have issues with numbered lists:
> https://antennapod.org/documentation/general/synchronization

What's the issue with the source/English page with the numbered lists? It seems that the list as such is fine, the indent is just a bit small.

ByteHamster:
The numbers are off when there are images in the list. It's 1,2,1,1,2,3,4

keunes:
Oh, doh, thanks. The source file seems fine, but I things things go wrong when Jekyll compiles stuff due to the image include or the 'variable' declaration. I'll have a look, but I'm fearing it'll be kinda hard to fix this :/

ByteHamster:
I mean, we could just replace it with an unordered list, then we don't have problems with the numbers not counting correctly ;)

loucasal:
I noticed that a number of tickets addressed by #460 are still open (unintentionally?).

In this case, though, a minor lingering issue remains:

> A page where the source page is displayed correctly, but localised pages have issues with bullet point/numbered lists:
https://antennapod.org/fr/documentation/bugs-first-aid/database-error

At https://antennapod.org/documentation/bugs-first-aid/database-error, the bullet point list is indented under item 3. In localised pages such as the one linked above, the indenting is lost.

ByteHamster:
I think the ones that are still open were not fully implemented yet

Maybe the formatting in markdown will get better once we move to weblate managed markdown instead of our own conversion scripts

Issue #314, opened 2024-01-07
Display cover image on subscribe page


keunes:
**Short description**: Display a feed's cover image when cover image URL is provided in a parameter.

**Location**: Subscription page: https://antennapod.org/deeplink/subscribe/?url=https://podcast.thelinuxexp.com/@tlenewspodcast/feed.xml&title=Linux%20%26%20Open%20Source%20News
I would display it left from the page/podcast title.

**Why have this**: Makes the subscription page more appealing

**More info**: 

* Builds on: https://github.com/AntennaPod/antennapod.github.io/pull/303
* Context: https://forum.antennapod.org/t/webplayer-share-landing-page-deeplink/3865
* Inspiration: https://podcastindex.org/podcast/5948837 / https://github.com/Podcastindex-org/web-ui/blob/master/ui/src/components/PodcastHeader/index.tsx#L94
* Need to check CORS. As noted by @ueen:
> CrossOriginResourceSharing would allow embedding images and audio. Whether this is supported by most podcasthosts is an open question that needs to be investigated.*
>
> *check this out [Podcastindex.org](https://podcastindex.org/podcast/1037491?episode=16729419694) image and audio doesnt seem to be cached, its linked straight from podigee-cdn

Nice to haves: display also 'author' and 'description'. On app side we need to consider [URL maximum length](https://www.baeldung.com/cs/max-url-length) handled by browsers - i.e. probably make the Description (stripped of any HTML) the last parameter and cut off at 2048 characters.

ueen:
Better: do a db with API and short URLs so not to encode in URL and have unhanable URLs, there for sure already an open source tool for this or smth very similar (short links)

Further improvement: embed mp3 like podcastindex to have a webplayer

ByteHamster:
If we actually commit on adding a db, do we want to make it just a redirect to the static website we currently have? That would mean that clicking the short links brings users to the long url, so they see it and could maybe even share the long url using their web browser. I think if we have a database anyway, the episode page could also maybe be managed as a non-static site. But that probably requires quite a bit of re-strucuturing of the server architecture

ueen:
No if where doing it anyways we should do an API with the app sending the info as json, saved to db and the landing page then getting that info by the short URL. It could be totally separate from everything else. I'm not an experienced dev and I've done similar small scale things like this in a day with SQL and php and js, just designing front end (CSS) is not my strong suit and I would like to not do that. For the rest I could code something simple. :)

[...]

Issue #279, opened 2023-10-14
Relative (internal) links should lead to translated versions


keunes:
**Short description**: Make relative/internal links send to their localised versions.

**Location**: https://antennapod.org/es/documentation/getting-started

**Why have this**: On the 'Getting started' page there's a relative/internal link to to the 'subscribe to a podcast' page. However, it sends folks to the English version of the page.

**More info**:

Issue #274, opened 2023-09-17
Automatic language detection


BaumiCoder:
**Short description**: 
If I visit the website white a non english browser, e.g. a german one, the website should automatically switch to this language. If the language isn't available on the website, english should be the default.

**Location**: 
Everywhere on the website. 

A special site is the subscribing page (documented [here](https://antennapod.org/documentation/podcasters-hosters/add-on-antennapod)). A manual switch of the language does remove the `url` parameter, a automatic switch should avoid that.

**Why have this**: 
Better accessibility for people without English skills, especially on the subscribing page.

PR #460, opened 2026-01-07
Major website improvements and new features implemented

### LINKED ISSUE #438: Code blocks are not adapted to dark mode
**Short description**:
Code blocks are displayed as dark text on dark background and are therefore very difficult to read.

**URL**: 
See the [automation](https://antennapod.org/documentation/automation/tasker) documentation for an example.

AnXh3L0:
### Major Changes

**Bootstrap 5 Upgrade**

Upgraded from Bootstrap 4 to Bootstrap 5.3. This removes the jQuery dependency entirely, adds native dark mode support via `data-bs-theme`, and brings modern CSS features. All components have been updated to use the new Bootstrap 5 classes and JavaScript APIs.

**Brand Color Unification**

The website was using Bootstrap's default blue in various places. This has been fixed - all buttons, badges, accordions, hover states, and focus rings now use the AntennaPod brand palette consistently. Dark mode has been improved with proper contrast ratios. A theme switcher (system/light/dark) has been added to the header.

---

### New Features

**New Pages**

- Features page (`/features`) - A detailed showcase of what the app can do, organized in an accordion layout by category
- Media kit page (`/media-kit`) - Brand assets for press and community use: logos in various formats, launcher icons, color palette with hex codes

**New Homepage Components**

- Interactive demo with app screenshots in a carousel
- Contributors card showing the top 56 GitHub contributors
- What's New card displaying recent releases
- Contributions graph showing commit activity
- Latest forum discussions (fetched from forum.antennapod.org with fallback proxies for CORS)
- Language showcase highlighting translation coverage

**Blog Improvements**

- Responsive thumbnails: a script generates 400px-wide versions of blog images for the overview page, significantly improving load times
- Category filtering has been rewritten to work properly with View Transitions
- Related posts section at the bottom of each article, based on matching categories
- Previous/next navigation to browse between posts

**Internationalization**

- Auto language detection: visitors are redirected to the localized version of the site based on their browser language (falls back to English)
- Link localization: internal links within non-English blog posts now correctly point to the localized destination pages
- Added translation keys for all new components (using English as placeholder - the community can translate these)

**Events Page**

On mobile, the "Meet us!" call-to-action block now appears between upcoming and past events, so users see concrete upcoming events before the generic CTA.

---

### Technical Changes

**Scripts**

- `scripts/generate-thumbnails.sh` - Bash script to generate blog image thumbnails using ImageMagick
- Jekyll plugins for fetching GitHub release data and Weblate translation statistics

**Assets**

- FontAwesome updated to latest version, removed TTF fonts (WOFF2 only)
- Dynamic SVG logo that switches colors based on the active theme
- Branding assets stored locally (logos, icons, color palette image)

**Cleanup**

- Updated `.gitignore` with common dev artifacts (.DS_Store, IDE files, etc.)
- Removed deprecated Bootstrap 4 files and the bootstrap-dark override folder
- Removed jQuery

---

While these changes have been fairly tested locally on my side, the localizations keys need to be translated into the other languages we support, and then the GitHub Actions probably has to be changed/edited in order to support the Jekyll plugins I added, but I am not 100% sure if it can detect and use them as-is. I would need confirmation from @ByteHamster or @keunes.

--- 

This PR should theoretically cover and close #438 #337 #188 #238 #62 #314 #203 #279 #274 #207 #189 #99 #111 #110 #100 #104 #103 #101 #98 #97


ByteHamster:
Wow, this is a huge improvement!

Some thoughts (which I did not discuss with @keunes yet)
- Do we need both the "Your Podcasts, Your Way" section and the "Ad-Free" / "Listen anywhere" section? Maybe just the features are enough?
- How do the live sections work (forum, GitHub)? Do we have to re-deploy the website every time we want it to be updated?
- I'm not sure about the cards taking more horizontal space than the text. Maybe the text could be a bit wider?
  grafik
- Should we remove the white space around the screenshots?
  grafik
- Should we use normal screenshots or the "framed" screenshots with text+background there?
- Maybe the contributors box that is currently on the downloads page could also be on the contribute page? Or maybe on the front page below the GitHub statistics? We recently discussed that we want to stress the developer community a bit more than on the old website, so this could improve it even more
- The features page with the collapsed sections reminds me of our [old FAQ page from many years ago](https://htmlpreview.github.io/?https://github.com/AntennaPod/antennapod.github.io/blob/cd2b34936742ef8131c55b3403b0489e687c934e/faq.html). How about expanding them all and making them independent cards? I like the idea of linking to the docs there!
- In the documentation search, the title is displayed twice: Once in the title and once prepended to the text
- On the front page, what does "4 this week" mean? The commits? Maybe that could be a bit more clear?
- The forum section on the front page says "just now" but the post was 2 days ago. What qualifies a post as "just now"?

(the new blog post interface looks nice! I should really write a post again to bring it more to life :D)

AnXh3L0:
@ByteHamster below answers to your points! Let me know if there's anything else you would like to discuss.

>	Do we need both the "Your Podcasts, Your Way" section and the "Ad-Free" / "Listen anywhere" section? Maybe just the features are enough?

Not necessarily, even one can be enough, if we decide on which one is better.

>	How do the live sections work (forum, GitHub)? Do we have to re-deploy the website every time we want it to be updated?

Currently some of the items are fetched on page build, them being the languages translation progress from Weblate and the GitHub releases changelogs. The other components are fetched on page load, and some of them are cached on the client side.

>	I'm not sure about the cards taking more horizontal space than the text. Maybe the text could be a bit wider?
>	grafik

The cards container could be made same width as the text, as the text container was previously too wide, making it hard to read on big screens.

>	Should we remove the white space around the screenshots?
>	grafik

Yes, we should remove the whitespace around the screenshots.

>	Should we use normal screenshots or the "framed" screenshots with text+background there?

I would suggest using normal screenshots, not framed ones if possible.

>	Maybe the contributors box that is currently on the downloads page could also be on the contribute page? Or maybe on the front page below the GitHub statistics? We recently discussed that we want to stress the developer community a bit more than on the old website, so this could improve it even more.

The contributors box can be used on the contribute page and even on the homepage, not an issue.

>	The features page with the collapsed sections reminds me of our [old FAQ page from many years ago](https://htmlpreview.github.io/?https://github.com/AntennaPod/antennapod.github.io/blob/cd2b34936742ef8131c55b3403b0489e687c934e/faq.html). How about expanding them all and making them independent cards? I like the idea of linking to the docs there!

Yes, good idea, we can turn them into cards as well, my only worry is that there might be too many cards at any point, if all features are inserted.

>	In the documentation search, the title is displayed twice: Once in the title and once prepended to the text

Will fix that, very small bug.

>	On the front page, what does "4 this week" mean? The commits? Maybe that could be a bit more clear?

Yes, it means 4 commits this week, we can improve the wording/clarity there.
 
>	The forum section on the front page says "just now" but the post was 2 days ago. What qualifies a post as "just now"?

I don't know if you're using the Netlify Preview to see the build or trying it locally, but it sometimes depends on the device, network, proxy etc, so we will have to check if there's an easy solution for this one.

ByteHamster:
> Not necessarily, even one can be enough, if we decide on which one is better.

An advantage of just the features would be that we would have 3 sections in total, which works well with the backgrounds :) Currently there is an area with changed background but without content towards the bottom of the page.

> Currently some of the items are fetched on page build, them being the languages translation progress from Weblate and the GitHub releases changelogs.

I think it would be good to have this automated somehow or client-side (probably easier?). Otherwise we have to remember one more thing during releases. Do you think doing this client-side is a problem?

> some of them are cached on the client side.

Ah, that explains why I didn't see a request in the developer tools :) I already wondered how that works

Now that I know that, I see that it uses corsproxy.io. If possible, I would try to reduce the dependencies on external services as much as possible. For the production website it should be easy to set a CORS policy on the forum, but I'm unsure about the local development version and the netlify previews. The previews are all on different subdomains, so we would have to whitelist something like `https://*.netlify.app/`, which could be risky. Apparently the discourse forum does not support setting the CORS headers only for specific "safe" routes

> The cards container could be made same width as the text, as the text container was previously too wide, making it hard to read on big screens.

I still feel like the text column is a bit narrow, but as long as text and cards match, I'm perfectly fine :)

> I don't know if you're using the Netlify Preview to see the build or trying it locally, but it sometimes depends on the device, network, proxy etc, so we will have to check if there's an easy solution for this one.

Yes I used Netlify. Turns out, the very first time I load the page (in a private tab), the posts show the correct time. As soon as I refresh, it turns into "just now" and stays that way. So this seems to be related to caching. If it simplifies the code to not cache, do we actually need the caching there?

AnXh3L0:
> > Not necessarily, even one can be enough, if we decide on which one is better.
> 
> An advantage of just the features would be that we would have 3 sections in total, which works well with the backgrounds :) Currently there is an area with changed background but without content towards the bottom of the page.

Yes, we can decide on unifying the sections and keeping just the features one for example.

> > Currently some of the items are fetched on page build, them being the languages translation progress from Weblate and the GitHub releases changelogs.
> 
> I think it would be good to have this automated somehow or client-side (probably easier?). Otherwise we have to remember one more thing during releases. Do you think doing this client-side is a problem?

The data fetching and caching is theoretically automated already, on build time, data is automatically fetched via the two Jekyll plugins I wrote, and then it's saved on the output build, there is no need to remember to do that yourself. I will check if it's viable to have them all fetch client-side on page load, and somehow cache for the current session, so that it won't keep hiting their APIs all the time.

> > some of them are cached on the client side.
> 
> Ah, that explains why I didn't see a request in the developer tools :) I already wondered how that works

Yes, some data is cached because there's usually no need to refetch on every single load.

> Now that I know that, I see that it uses corsproxy.io. If possible, I would try to reduce the dependencies on external services as much as possible. For the production website it should be easy to set a CORS policy on the forum, but I'm unsure about the local development version and the netlify previews. The previews are all on different subdomains, so we would have to whitelist something like `https://*.netlify.app/`, which could be risky. Apparently the discourse forum does not support setting the CORS headers only for specific "safe" routes

Regarding the proxy, yes, I would like to find a solution that doesn't involve third parties at all, but in this case the proxy is necessary for the CORS headers, so that data can work, otherwise, we would end up with having to build the website at least daily, in order to have more or less up-to-date information for the forum posts.

> > The cards container could be made same width as the text, as the text container was previously too wide, making it hard to read on big screens.
> 
> I still feel like the text column is a bit narrow, but as long as text and cards match, I'm perfectly fine :)

Yes, I will match the widths for both.

> > I don't know if you're using the Netlify Preview to see the build or trying it locally, but it sometimes depends on the device, network, proxy etc, so we will have to check if there's an easy solution for this one.
> 
> Yes I used Netlify. Turns out, the very first time I load the page (in a private tab), the posts show the correct time. As soon as I refresh, it turns into "just now" and stays that way. So this seems to be related to caching. If it simplifies the code to not cache, do we actually need the caching there?

Is it indeed relating to the caching, which I will check again.

ByteHamster:
> I would like to find a solution that doesn't involve third parties at all, but in this case the proxy is necessary for the CORS headers

Would you mind trying this url instead? :) https://corsproxy.forum.antennapod.org/latest.php

PR #476, opened 2026-02-15
Re-enable forum posts section

### LINKED ISSUE #100: 'Latest form posts' list
**Short description**: Display a block with the 5 forum threads most recently updated, with the title, category, and possibly also last activity date (similar to what Discourse displays). Each entry is a link that opens the respective forum thread.

**Location**: Homepage

**Why have this**: Show that we have an active community; encourage engagement in that community  (so that in the end we have more people helping the project with user support).

**More info**: 
* With a static site, I imagine we want to get this only when a visitor actually might view it, so we should [load this block lazily](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading) (only when user scrolls down).
* It could be similar to what's visible on Discourse itself (though without the number of posts):
(forum header below only in the screenshot because Firefox included it and I'm too lazy to crop it off)
afbeelding

ByteHamster:
Fixed caching and connecting to forum server directly instead of relying on third-party CORS proxy

Closes #100

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -99,9 +99,9 @@ homepage:
       description: "With AntennaPod, you can download episodes and listen anywhere. You can even select podcasts that should be downloaded automatically."
       img-alt:
   forum-posts:
-    title: "Community Forum"
+    title: "Community forum"
     description: "Get help, share ideas, and connect with other users"
-    view-all: "Visit the Forum"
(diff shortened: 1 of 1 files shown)


ByteHamster:
There will be a section about GitHub next to it soon, so I wouldn't make it full width

PR #496, opened 2026-05-01
Re-enable contributions widget on website

### LINKED ISSUE #101: Code contributions graph
**Short description**: A simple graph showing code contributions over time (number of commits per week, over the past 6 months). There should be a call 'Join development' or 'Contribute to AntennaPod' in close proximity.

**Location**: Homepage, bit further down.

**Why have this**: Showcase that the project is under active development, and encourage people to contribute, too.

**More info**: 
* Could look similar to the contributor graph on GitHub:
afbeelding
* Another layer could be added, indicating the number of contributors during that period (the number of accounts/people that submitted those commits). Though that might make it a bit too complex.
* It would be more interesting (than number of contributors) to have another layer in the graph for the number of strings that were translated (where each language counts as one), to consider & display also forms of contributions other than code.

ByteHamster:
Was disabled on the new website because it was not ready yet. Now it is done and can be enabled. Also unifies the caching logic to avoid code duplication.

Closes #101

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -361,8 +361,8 @@ events:
 #      description: "Set up automatic downloads, deletions, and podcast refreshes. Integrate with Tasker for even more powerful automation."
 
 ## Contributors
-#contributors:
-#  title: "Contributors"
+contributors:
+  title: "Contributors"
 #  description: "AntennaPod is built by amazing volunteers from around the world."
 #  view-all: "View all on GitHub"
 #  loading: "Loading contributors..."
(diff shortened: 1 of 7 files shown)


keunes:
Nice! I would make some design tweaks to align it with the forum widget. I can take a look when I'm back home.

ByteHamster:
I think it already looks nice and consistent but feel free to push some tweaks :)

ByteHamster:
@keunes should we merge this version already and then do the tweaks later? I mean, it is what the designer created

keunes:
I worked on a slightly tweaked design (well, got Copilot to do it).

Before:
image

After:
image

Any objections to this change @ByteHamster?

<details><summary>Alternative with aligned header but keeping the CTA in the header</summary>
image
</details>

ByteHamster:
No objections

PR #499, opened 2026-05-10
Replace ICS file with calendar links and clean up

keunes:
Update the event system to replace the ICS file with direct links to add events to calendars.


keunes (commented):
To check: possibly update 'Community meeting' to 'Community call'; use page slug as default for recurring meeting calendar URLs; make feature request against Davis.

keunes:
@ByteHamster on this point:
> use page slug as default for recurring meeting calendar URLs

What do you think? It is definitely feasible. We can also get rid of the 'meeting room' front matter using similar logic. Both (`calendar-path` and `meeting-room`) can be set automatically based on the slug (file name) if the meeting/event is online and a) recurring or b) one-off and upcoming.

The downside is that it's less explicit:
* In (the unlikely) case we have another recurring event, we need to remember to set up the redirect in the htaccess file with the same slug as the event file on the website.
* For all new (one-off and recurring) events we need to remember to align the meeting room URL that we add in the calendar with the slug as the event file on the website.

I think that this is fine, though. In any case I would make these front matter fields optional, so we can still explicitly set them to something different if we want.

ByteHamster:
If it is unlikely that we add another recurring event, does it matter to have it filled in automatically? Making it explicit means that we need to remember less. An AI won't remember this either. If it's explicit in the code, there is less risk for things going wrong

keunes (commented):
Ok, thanks, makes sense - let's leave it explicit. Then the PR is ready to review.
Feature request against Davis to have event-specific subscription URLs: https://github.com/tchapi/davis/issues/260

_data/event-links.yml
```diff
@@ -1,16 +1,16 @@
 # used for event layout
-- key: ICS
+- key: calendar-path
   icon: calendar-plus
   text: events.links.add-calendar
-  url: /events.ics
+  url: https://calendar.antennapod.org/
```
ByteHamster, line 5: This should be updated, I assume?

_events/0_A meeting template.md
```diff
@@ -1,49 +1,37 @@
...
 blog-post: 
 # Optional. Must be the path of the blog post, starting with the year (e.g. 2022/10/translation-coordinator).
+recording:
+# Optional. Must be a URL.
+
+# Calendar links:
+calendar-path:
+# Required for recurring events. Path under https://calendar.antennapod.org, for example "community-call".
+uid:
+# Required for one-off events. Davis/Sabre event UID used to build the direct .ics download URL.
```
ByteHamster, line 33: Can't we just use the same calendar link property for both instead of trying to be smart? That's a lot more explicit

_events/community-call.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Monthly community call ☎ 🎙
+summary: Our monthly call to discuss anything AntennaPod (and beyond). Join us!
+uid: 0718779a-5b92-48df-87e3-801e02154c06
```
ByteHamster, line 4: Why do we have both uuid and path here?

_i18n/en/general/events-add-calendar-modal.md
```diff
@@ -1,3 +1 @@
-Always want to have the latest community call dates and times in your calendar? Then copy the URL and add it as a remote calendar in your calendar app.
-
-Hope to see you!
+Want to download the calendar file of this gathering, or subscribe to all our calls and meetings?
```
ByteHamster: Gathering -> Event? That's what we call it in the menu

ByteHamster: I think the explanation that you can add the url to the calendar app was useful. I assume not many people know what to do with a calendar link, and without the text they don't really have a pointer where to start

_plugins/recurring_events_permalink.rb
```diff
@@ -0,0 +1,17 @@
+module Jekyll
```
ByteHamster, line 2: I don't understand why we need custom code here. Don't we get those links automatically? On the old site we had the same links without needing ruby code

_data/event-links.yml
keunes: No, it's the baseline URL for our calendar links. (E.g. https://calendar.antennapod.org/all-events gets created from it.)

_events/0_A meeting template.md
keunes: The logic is:
* recurring events: we want to encourage people to subscribe to the remote calendar and receive updates (whether they subscribe to all events or just this one)
* one-off events: if they don't want to subscribe to all our events, we should offer a quick way to add just this one (e.g. FOSDEM) to their calendar.

For this distinction, we need different info. We create human-readable slugs via htaccess in Davis for recurring events, which we shouldn't be bothered to do for one-off events.

Side note: UID is actually required always - our PR workflow fails without it.
* For now: Also for recurring events we offer the same options as for one-off events, until https://github.com/tchapi/davis/issues/260 might get addressed
* Long-term: While not strictly necessary for recurring event links, I think it's still good to mandate it just in case. I'll update the description in the event template file.

_i18n/en/general/events-add-calendar-modal.md
keunes: My logic was that we already hint visitors to this option right below the link button (which, granted, many might not read):
image

But yes, we can add back the explanation.
```suggestion
Copy the URL and add it as a remote calendar in your calendar app to add all our calls and meetings. Or download the ICS file to add just this event.
```

_plugins/recurring_events_permalink.rb
keunes: Event links by default (as per `_config.yml`) get a year and month in their slug. This doesn't make sense for recurring events. That's why in the past we manually overrode the permalink for recurring events. I replaced that with this script (which sets the page slug based on `calendar-path`) to ensure consistency between the different URLs (online meeting room, Jekyll page slug, Davis short link) without us (or future contributors) having to think about/remember this.

See 9125a81bec85687d2e903e23ffcb4a135bad4fc9

_i18n/en/general/events-add-calendar-modal.md
keunes: We might even decide to remove that 'Always want to have the most up-to-date...' block.

ByteHamster: Yeah, I think I would rather remove that block and show it directly where you copy the link. This should make it less likely for users to miss, but they only see it if they need it.

_plugins/recurring_events_permalink.rb
ByteHamster: That makes it less obvious what is happening, though. Not 100% convinced. At the very least the file could get a short comment explaining the script and the reasoning behind it

_events/0_A meeting template.md
ByteHamster: I think the GUID was only needed when we generated events manually. Now, with the dedicated calendar tool, I think we can remove the CI check and the GUID from the files. Then we can still give the path to whatever Davis ends up giving as link.

keunes: Ok, we can drop the GUID being mandatory for events. Events then should have either a `calendar-path` or a `guid` (as written in the template).
Because it's currently not possible to subscribe to a specific event via CalDAV (see link to Davis), we have two options for recurring events:

1. provide only an `all-events` CalDAV link
   * downside: folks are maybe less likely to subscribe as they're only interested in the community call and don't want to have all the Needs: Decision calls.
2. provide an `all-events` CalDAV link + `guid`-based iCal link to the event series
   * downside: the iCal link doesn't provide calendar name or colour and (at least on Android) you need a special app (e.g. [ICSx<sup>5</sup>](https://icsx5.bitfire.at/)) to subscribe.

<details>
<summary>difference between CalDAV and iCal according to Claude</summary>

> Both rely on polling — neither method has true server push in practice.
>
> With **iCal**, the client downloads the full `.ics` file on whatever interval it's configured for (often every hour or even less frequently, since it's a crude fetch).
>
> With **CalDAV**, it's also polling, but typically at a shorter interval (Thunderbird defaults to every 30 minutes, though configurable). The advantage is that CalDAV only fetches what's changed since the last sync rather than re-downloading everything, so shorter poll intervals are cheaper and more practical.
>
> There is a CalDAV extension called **calendar-auto-schedule** and a push notification mechanism, but real server-push for CalDAV is rare in self-hosted setups and Sabre/Davis doesn't implement it as far as I know. So in your case, both methods would effectively behave the same way in terms of update latency — it's just a matter of how aggressively the client polls.
</details>

Do you have a preference between the two?

ByteHamster: 2 sounds nice to me. I mean, for CalDav you also need an app right?

Issue #513, opened 2026-05-29
Add goPodder to the synchronization docs?


cbrgm:
The [sync docs](https://antennapod.org/documentation/general/synchronization) list Nextcloud, gPodder server, and Micro GPodder server. I maintain another gPodder-compatible server called goPodder (https://github.com/cbrgm/gopodder) and have been using it with AntennaPod without issues.

Would you be open to adding it to the list? Happy to open a PR

ByteHamster:
We should probably reword that part of the docs a bit and give a list of servers and their programming language. Currently the list sounds a bit weird, especially when translated. But I'm very open to adding more implementations :) would you mind opening a PR? 

Some others that I remember:
- https://github.com/kd2org/opodsync
- https://github.com/bobrippling/podsync
- https://github.com/eliassoares/malipod-selfhosted

Maybe we should also add a sentence that those are 3rd party projects not affiliated with AntennaPod

cbrgm:
Yes, sounds good to me. I will assign this issue to me, then provide a PR. Thanks for the feedback!
Edit: Oh wait, I can't actually assign it to me. Never mind :)

PR #514, opened 2026-05-29
docs: list self-hosted sync servers with programming languages

cbrgm:
Addresses #513.

Rewords the self-hosted sync section into a list format showing each server with its programming language, as suggested by @ByteHamster. Adds opodsync, podsync, malipod, and goPodder. Includes a note that these are third-party projects not affiliated with AntennaPod.


cbrgm:
Heads up: I initially also updated the German translation but reverted it after noticing that translations are handled by a bot. Only the English file is changed now.

ByteHamster:
Thanks!

